PR #5, opened 2023-02-24
Use package resolution from JavaHelpers.java

buildbreaker:
Bringing in `JavaHelper.java` methods from https://codereview.appspot.com/912042

Fixes:
- `java_multiple_files` option interaction with Connect-Kotlin's generator


protoc-gen-connect-kotlin/src/main/kotlin/build/buf/protocgen/connect/internal/ProtoHelpers.kt
```diff
@@ -0,0 +1,476 @@
...
+ * Several code generators treat the parameter argument as holding a list of
+ * options separated by commas: e.g., `"foo=bar,baz,qux=corge"` parses
+ * to the pairs: `("foo", "bar"), ("baz", ""), ("qux", "corge")`.
+ *
+ *
+ * When a key is present several times, only the last value is retained.
+ */
+internal fun parseGeneratorParameter(
+    text: String
+): Map<String, String> {
```
jhump, line 79: I think you've written something that is redundant with this. So you could probably just use this version (or replace this with what you wrote).

protoc-gen-connect-kotlin/src/main/kotlin/build/buf/protocgen/connect/internal/ProtoHelpers.kt
```diff
@@ -0,0 +1,476 @@
...
+
+/**
+ * Commonly-used separator comments (a line of '=').
+ */
+internal const val THICK_SEPARATOR = "// ===================================================================\n"
+
+/**
+ * Commonly-used separator comments (a line of '-').
+ */
+internal const val THIN_SEPARATOR = "// -------------------------------------------------------------------\n"
```
jhump, line 64: Can probably remove these 🤷

protoc-gen-connect-kotlin/src/main/kotlin/build/buf/protocgen/connect/Generator.kt
```diff
@@ -46,27 +48,20 @@ import com.squareup.kotlinpoet.asClassName
  * move off of type aliases, this can be changed without user API breakage.
  */
 private val HEADERS_CLASS_NAME = ClassName.bestGuess("build.buf.connect.Headers")
```
jhump, line 50: FYI, can also do `ClassName.get("build.buf.connect", "Headers")` to directly indicate the package and the class name (instead of making `bestGuess` figure it out via string parsing).

protoc-gen-connect-kotlin/src/main/kotlin/build/buf/protocgen/connect/Generator.kt
```diff
@@ -106,28 +101,28 @@ class Generator : CodeGenerator {
...
     ): List<FunSpec> {
         val functions = mutableListOf<FunSpec>()
         val headerParameterSpec = ParameterSpec.builder("headers", HEADERS_CLASS_NAME)
             .defaultValue("%L", "emptyMap()")
             .build()
         for (method in methods) {
-            val inputClassName = ClassName.bestGuess(className(method.inputType))
-            val outputClassName = ClassName.bestGuess(className(method.outputType))
-            if (method.hasClientStreaming() && method.hasServerStreaming()) {
+            val inputClassName = ClassName.bestGuess(getClassName(method.inputType))
```
jhump, line 123: I think we'll want to create a helper for this to actually compute the correct class name, instead of using `bestGuess`. Otherwise, if someone is not using expected naming conventions, this could return an incorrect name.

I don't think it's hard either. You basically get the package for the given element's file as the first component. Then you strip that package prefix from the element's full name. And then you split what remains on `.`.  Then you have the package and the sequence of (possibly enclosing) class names that you can supply to `ClassName.get`.

buildbreaker: Sounds good. I think I was able to pull the right information with this:
```
  private fun classNameFromType(inputType: Descriptors.Descriptor): ClassName {
      if (inputType.file.options.javaMultipleFiles) {
          return ClassName(getFileJavaPackage(inputType.file), inputType.name)
      }
      return ClassName(getFileJavaPackage(inputType.file), getFileClassName(inputType.file), inputType.name)
  }
```

protoc-gen-connect-kotlin/src/test/kotlin/PluginGenerationTest.kt
```diff
@@ -0,0 +1,49 @@
...
+    @Test
+    fun emptyPackagePresence() {
+        val request = NoPackage.SayRequest.newBuilder().setSentence("hello").build()
+        assertThat(request.sentence).isEqualTo("hello")
+    }
+
+    @Test
+    fun multiFileEnabled() {
+        assertThat(EnabledEmptyServiceClient::class.java).isNotNull
+        assertThat(EnabledServiceClient::class.java).isNotNull
```
buildbreaker, line 35: I'm not fully sure how to appropriately these naming type changes but this assumes that the underlying service itself would be present and can be expected to compile

protoc-gen-connect-kotlin/src/main/kotlin/build/buf/protocgen/connect/Generator.kt
jhump, line 50:
```suggestion
private val HEADERS_CLASS_NAME = ClassName("build.buf.connect", "Headers")
```

protoc-gen-connect-kotlin/src/test/kotlin/PluginGenerationTest.kt
jhump: I think for now, making sure the generated code is compilable is likely sufficient.

protoc-gen-connect-kotlin/src/main/kotlin/build/buf/protocgen/connect/Generator.kt
```diff
@@ -310,21 +306,19 @@ class Generator : CodeGenerator {
...
-        val className = type.split(".").last()
-        if (compiledTypes.contains(className)) {
-            val packageName = compiledTypes[className]
-            return "$packageName.$className"
+    private fun classNameFromType(inputType: Descriptors.Descriptor): ClassName {
+        if (inputType.file.options.javaMultipleFiles) {
+            return ClassName(getFileJavaPackage(inputType.file), inputType.name)
         }
-        return type.substring(1)
+        return ClassName(getFileJavaPackage(inputType.file), getFileClassName(inputType.file), inputType.name)
```
jhump, line 313: This isn't sufficient. It will generate an incorrect name for a message that is defined _inside of_ another message.

I was thinking of something more like so:
```suggestion
        val packageName = getFileJavaPackage(inputType.File)
        val names = getClassName(inputType).
                removePrefix(packageName).
                removePrefix(".").
                split(".").
                toTypedArray()
        return ClassName(packageName, names[0], *names.slice(1..names.size-1))
```

buildbreaker: You're right. I added another proto file for regression prevention

jhump (approved):
LGTM

PR #10, opened 2023-02-28
Add examples for callbacks in simple main

buildbreaker:
Adding the callback examples to main function examples. The intention here is to provide a straightforward example for some of our friends who want to use the callback signature.

### DIFF
--- a/examples/buf.gen.yaml
+++ b/examples/buf.gen.yaml
@@ -5,6 +5,8 @@ managed:
 plugins:
   - name: connect-kotlin
     out: generated-google-java/src/main/kotlin
+    opt:
+        - generateCallbackMethods=true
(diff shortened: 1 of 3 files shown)


rebello95 (commented):
What does this do when running the actual demo? Send "hello" twice in a row, I assume?

buildbreaker:
Yeah it sends hello twice. I can make a switch statement that reads the user input but seeing as how this example is primarily to illustrate, I decided to avoid that. Do you think we should do something more sophisticated?

rebello95:
Yeah I agree that a switch is probably unnecessary. Just seems a little weird to sacrifice the behavior of the actual app for example code -- maybe just do something like `// Alternatively, callbacks can be used. For example:`? Fwiw connect-swift only has an async/await demo 🤷🏽

Issue #12, opened 2023-03-02
Base URL has no validation


kohenkatz:
I noticed that my Android app was using URLs with a double slash after the hostname, like this:

```
https://server.example.com//com.example.v1.ExampleService/ExampleEndpoint
```

I realized that this is happening because I store my server base URL as an `okhttp3.HttpUrl` object, and then create the `ProtocolClientConfig` like this:

```kotlin
fun createConnectClient(): ProtocolClientInterface {
    return ProtocolClient(
        httpClient = ConnectOkHttpClient(),
        ProtocolClientConfig(
            host = server.toString(),
            serializationStrategy = GoogleJavaLiteProtobufStrategy(),
            networkProtocol = NetworkProtocol.CONNECT,
            interceptors = listOf(
                { AuthenticationInterceptor(User.bearerToken) },
                { LoggingInterceptor() },
            )
        ),
    )
}
```

For now, I work around this issue by manually building the host URL string, like this:

```kotlin
fun createConnectClient(): ProtocolClientInterface {
    return ProtocolClient(
        httpClient = ConnectOkHttpClient(),
        ProtocolClientConfig(
            // We cannot use `server.toString()` because it adds a trailing slash
            host = "${server.scheme}://${server.host}:${server.port}",
            serializationStrategy = GoogleJavaLiteProtobufStrategy(),
            networkProtocol = NetworkProtocol.CONNECT,
            interceptors = listOf(
                { AuthenticationInterceptor(User.bearerToken) },
                { LoggingInterceptor() },
            )
        ),
    )
}
```

However, I do not think this is a good long-term solution, for three reasons:

1. It is confusing for future developers who read the code and need to understand why we did this (requiring extra comments to explain).
2. In general, it would be nice to have some kind of validation that a proper URL has been provided.
3. The way that `ProtocolClient` builds a URL by concatenating strings (`url = URL("${config.host}/${methodSpec.path}")`) seems fragile.

I suggest that the `host` be stored as a `URL` object instead of as a `String`. Doing that fixes all three of my issues listed above, as follows:

1. Constructing the URL object from a String doesn't care if there's a trailing slash or not (or, in my case, using the `toUrl()` method on the `HttpUrl` object).
2. If an invalid URL is provided, the exception will be thrown immediately on construction instead of waiting until execution to fail.
3. You can build new URLs without string concatenation using the two-argument URL constructor that takes a base URL as "context": `url = URL(config.host, methodSpec.path)`

---

[...]

PR #17, opened 2023-03-06
Ensure URL resiliency

buildbreaker:
Adding URI parsing resiliency by relying on `URI` to help parse the host and scheme. The previous implementation contained some brittleness when the user sets a host to be `https://example.com/` rather than `https://example.com`

Resolves https://github.com/bufbuild/connect-kotlin/issues/12

### DIFF
--- a/library/src/main/kotlin/build/buf/connect/impl/ProtocolClient.kt
+++ b/library/src/main/kotlin/build/buf/connect/impl/ProtocolClient.kt
@@ -30,7 +30,7 @@ import build.buf.connect.http.HTTPRequest
 import build.buf.connect.http.Stream
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.suspendCancellableCoroutine
-import okhttp3.HttpUrl.Companion.toHttpUrl
+import java.net.URI
 import java.net.URL
 import kotlin.coroutines.resume
 
@@ -53,9 +53,10 @@ class ProtocolClient(
         val serializationStrategy = config.serializationStrategy
         val requestCodec = serializationStrategy.codec(methodSpec.requestClass)
         try {
-            val httpURL = config.host.toHttpUrl().newBuilder().addPathSegment(methodSpec.path).build()
+            val host = URI(config.host)
+            val uri = URI(host.scheme, host.host, "/${methodSpec.path}", null)
             val unaryRequest = HTTPRequest(
-                url = httpURL.toUrl(),
+                url = uri.toURL(),
                 contentType = "application/${requestCodec.encodingName()}",
                 headers = headers,
                 message = requestCodec.serialize(request)
--- a/library/src/main/kotlin/build/buf/connect/impl/ProtocolClient.kt
+++ b/library/src/main/kotlin/build/buf/connect/impl/ProtocolClient.kt
@@ -203,8 +203,7 @@ class ProtocolClient(
     }
 
     private fun <Input : Any, Output : Any> urlFromMethodSpec(methodSpec: MethodSpec<Input, Output>): URL {
-        val host = URI(config.host)
-        val uri = URI(host.scheme, host.host, "/${methodSpec.path}", null)
-        return uri.toURL()
+        val host = URI(config.host).resolve(methodSpec.path)
+        return host.toURL()
     }
 }
--- a/library/src/main/kotlin/build/buf/connect/impl/ProtocolClient.kt
+++ b/library/src/main/kotlin/build/buf/connect/impl/ProtocolClient.kt
@@ -30,6 +30,7 @@ import build.buf.connect.http.HTTPRequest
 import build.buf.connect.http.Stream
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.suspendCancellableCoroutine
+import okhttp3.HttpUrl.Companion.toHttpUrl
 import java.net.URL
 import kotlin.coroutines.resume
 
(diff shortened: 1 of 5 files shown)


library/src/test/kotlin/build/buf/connect/impl/ProtocolClientTest.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.mockito.kotlin.any
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.whenever
+
+class ProtocolClientTest {
+    private val serializationStrategy: SerializationStrategy = mock { }
+    private val codec: Codec<String> = mock { }
+
+    @Test
+    fun urlConfiguration() {
```
rebello95, line 18: I don't see actual validation being done here - am I missing something?

rebello95 (commented):
Can you please also add some notes to the PR description / reference any relevant issues?

library/src/main/kotlin/build/buf/connect/impl/ProtocolClient.kt
```diff
@@ -52,8 +53,9 @@ class ProtocolClient(
         val serializationStrategy = config.serializationStrategy
         val requestCodec = serializationStrategy.codec(methodSpec.requestClass)
         try {
+            val httpURL = config.host.toHttpUrl().newBuilder().addPathSegment(methodSpec.path).build()
```
kohenkatz: The `config.host` value never changes, so why call `toHttpUrl` every time? Why not call it only once and store it so that you don't have the parsing overhead every time an endpoint is called?

library/src/test/kotlin/build/buf/connect/impl/ProtocolClientTest.kt
buildbreaker: The existing behavior fails depending if the user sets the "host" with a trailing `/` or not. This test exercises both those use cases

library/src/main/kotlin/build/buf/connect/impl/ProtocolClient.kt
buildbreaker: Yeah, that was probably not the right solution as it's an okhttp object

library/src/test/kotlin/build/buf/connect/impl/ProtocolClientTest.kt
buildbreaker: Done

library/src/main/kotlin/build/buf/connect/impl/ProtocolClient.kt
```diff
@@ -52,8 +53,10 @@ class ProtocolClient(
         val serializationStrategy = config.serializationStrategy
         val requestCodec = serializationStrategy.codec(methodSpec.requestClass)
         try {
+            val host = URI(config.host)
```
kohenkatz, line 56: This is still re-parsing the string every time you make a new request. Why do that more than once if the string cannot change?

kohenkatz:
You changed the `unary` method, but missed the same code in the `bidirectionalStream` method (line 144).

library/src/main/kotlin/build/buf/connect/impl/ProtocolClient.kt
```diff
@@ -200,4 +201,10 @@ class ProtocolClient(
             )
         )
     }
+
+    private fun <Input : Any, Output : Any> urlFromMethodSpec(methodSpec: MethodSpec<Input, Output>): URL {
+        val host = URI(config.host)
```
kohenkatz, line 206: Another point I raised in #12 was that the validation only happens when the `unary` or `stream` methods are called, instead of when the configuration object is originally constructed. I think that this makes it harder for a developer to discover that they've provided an invalid URL because they don't find out immediately.

For example, what happens if you do this?

```
fun createConnectClient(): ProtocolClientInterface {
    return ProtocolClient(
        httpClient = ConnectOkHttpClient(),
        ProtocolClientConfig(
            // We cannot use `server.toString()` because it adds a trailing slash
            host = "htp://example.com",
            serializationStrategy = GoogleJavaLiteProtobufStrategy(),
            networkProtocol = NetworkProtocol.CONNECT
        ),
    )
}
```

Note the `scheme` is `htp` instead of `http`. With the code as shown here, you only discover the typo later in the execution when you try to call the endpoint.

I think it would be helpful for the developer experience to know right away that an invalid value has been provided by doing this validation in the creation of the `config` object. Something like:

```
val baseUrl = URI(host)
```

Then you can use it here in `urlFromMethodSpec` with `config.baseUrl`.

buildbreaker: `URI()` doesn't actually do scheme validation. 

What I can do is to have a validation step on the configuration construction (which is basically `baseUri.toURL()`). If this is something that you're interested in for DX, I would like to separate that out in another PR and get connect-swift to also be consistent in behavior.

kohenkatz: I was thinking about okhttp's `toHttpUrl`, which I think does validation.

Also, I gave a bad example by giving a typo in the scheme as what we'd want validation for. A better example might be something that is completely malformed, which I think will fail creating a URI.

In any case, yes that can be a separate PR.

Thank you for looking at this.

buildbreaker: Oh yeah, this layer doesn't have the okhttp `toHttpUrl` dependency. We gotta use vanilla URI/URL

library/src/test/kotlin/build/buf/connect/ProtocolClientConfigTest.kt
```diff
@@ -0,0 +1,45 @@
...
+        val config = ProtocolClientConfig(
+            host = "https://connect.build",
+            serializationStrategy = mock { }
+        )
+        assertThat(config.baseUri.host).isEqualTo("connect.build")
+        assertThat(config.baseUri.toURL()).isNotNull()
+    }
+
+    @Test(expected = MalformedURLException::class)
+    fun unsupportedSchemeErrorsWhenTranslatingtoURL() {
```
rebello95, line 37:
```suggestion
    fun unsupportedSchemeErrorsWhenTranslatingToURL() {
```

rebello95 (approved):

PR #19, opened 2023-03-07
Update GitHub release job to include plugin artifacts

buildbreaker:
Resolves https://github.com/bufbuild/connect-kotlin/issues/4

Adding the protoc-gen-connect-kotlin jar to the releases.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,7 +42,7 @@ jobs:
           gpg --passphrase $GPG_PASSPHRASE --batch --import signing-key
           export ORG_GRADLE_PROJECT_signingInMemoryKey=$(gpg --armor --passphrase $GPG_PASSPHRASE --pinentry-mode=loopback --export-secret-keys $GPG_KEY_NAME signing-key | grep -v '\-\-' | grep -v '^=.' | tr -d '\n')
           shred signing-key
-          make buildplugin
+          make release
       - name: sha256
         run: |
           sha256() {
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -41,3 +43,13 @@ jobs:
           export ORG_GRADLE_PROJECT_signingInMemoryKey=$(gpg --armor --passphrase $GPG_PASSPHRASE --pinentry-mode=loopback --export-secret-keys $GPG_KEY_NAME signing-key | grep -v '\-\-' | grep -v '^=.' | tr -d '\n')
           shred signing-key
           make release
+      - name: sha256
+        run: sha256sum ./protoc-gen-connect-kotlin/build/libs/protoc-gen-connect-kotlin.jar >> sha256.txt
+      - name: Upload release artifacts in GitHub
```
rebello95, line 48: Might want to change this to something like `Publish release` since this will actually create a release

rebello95 (approved):

.github/workflows/release.yml
buildbreaker: Yeah there's two publishing going on here though: the sonatype/maven publish and this github one.

buildbreaker: Publish GitHub artifacts?

rebello95: Sure, the main thing I'm saying is that it creates a GitHub release too, and then pushes the artifacts with the release

PR #22, opened 2023-03-09
Update send() to forward all errors

buildbreaker:
Resolves https://github.com/bufbuild/connect-kotlin/issues/23

I'm working on getting some Android based unit tests working which should validate this behavior. However, I think it might incur some risk for our release tomorrow so I will have to do that work in a separate PR.

### DIFF
--- a/library/src/main/kotlin/build/buf/connect/http/HTTPClientInterface.kt
+++ b/library/src/main/kotlin/build/buf/connect/http/HTTPClientInterface.kt
@@ -56,8 +56,12 @@ class Stream(
         if (isClosed()) {
             return Result.failure(IllegalStateException("cannot send. underlying stream is closed"))
         }
-        onSend(buffer)
-        return Result.success(Unit)
+        return try {
+            onSend(buffer)
+            Result.success(Unit)
+        } catch (e: Throwable) {
+            Result.failure(e)
+        }
     }
 
     fun close() {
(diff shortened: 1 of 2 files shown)


library/src/main/kotlin/build/buf/connect/http/HTTPClientInterface.kt
```diff
@@ -56,8 +56,12 @@ class Stream(
         if (isClosed()) {
             return Result.failure(IllegalStateException("cannot send. underlying stream is closed"))
         }
-        onSend(buffer)
-        return Result.success(Unit)
+        return try {
```
buildbreaker, line 59: This is the contract with the underlying network -- we expect `onSend` to be unsafe and will handle that failure appropriately.

rebello95 (approved):

okhttp/src/main/kotlin/build/buf/connect/okhttp/OkHttpStream.kt
```diff
@@ -193,7 +193,7 @@ internal class PipeDuplexRequestBody(
                 bufferedSink.writeAll(buffer)
                 bufferedSink.flush()
             }
-        } catch (e: Throwable) {
+        } finally {
```
buildbreaker, line 196: Don't swallow these errors anymore and let the caller handle errors

PR #36, opened 2023-04-06
Fix request stream close on stream pause

buildbreaker:
Resolves https://github.com/bufbuild/connect-kotlin/issues/31


okhttp/src/main/kotlin/build/buf/connect/okhttp/OkHttpStream.kt
```diff
@@ -193,8 +193,9 @@ internal class PipeDuplexRequestBody(
                 bufferedSink.writeAll(buffer)
                 bufferedSink.flush()
             }
-        } finally {
+        } catch(e: Throwable) {
```
buildbreaker, line 196:
```suggestion
        } catch (e: Throwable) {
```

rebello95 (approved):

PR #39, opened 2023-04-26
Sanitize generated comments

stefanvanburen:
We're seeing issues with the codegen in which comments containing `*/` literals aren't being escaped. (An example can be found in [`google.longrunning.Operations.ListOperations`][1]).

KotlinPoet has an [open issue for sanitizing comments][2], so for now, borrow [their implementation in Wire][3].

[1]: https://bufbuild.internal/googleapis/googleapis/docs/main:google.longrunning#google.longrunning.Operations.ListOperations
[2]: https://github.com/square/kotlinpoet/issues/887
[3]: https://github.com/square/wire/pull/1445

### DIFF
--- a/protoc-gen-connect-kotlin/src/main/kotlin/build/buf/protocgen/connect/Generator.kt
+++ b/protoc-gen-connect-kotlin/src/main/kotlin/build/buf/protocgen/connect/Generator.kt
@@ -126,7 +126,7 @@ class Generator : CodeGenerator {
         val interfaceBuilder = TypeSpec.interfaceBuilder(serviceClientInterfaceClassName(packageName, service))
         val functionSpecs = interfaceMethods(service.methods, sourceInfo)
         return interfaceBuilder
-            .addKdoc(sourceInfo.comment())
+            .addKdoc(sourceInfo.comment().sanitizeKdoc())
             .addFunctions(functionSpecs)
             .build()
     }
@@ -144,7 +144,7 @@ class Generator : CodeGenerator {
             val outputClassName = classNameFromType(method.outputType)
             if (method.isClientStreaming && method.isServerStreaming) {
                 val streamingBuilder = FunSpec.builder(method.name.lowerCamelCase())
-                    .addKdoc(sourceInfo.comment())
+                    .addKdoc(sourceInfo.comment().sanitizeKdoc())
                     .addModifiers(KModifier.ABSTRACT)
                     .addModifiers(KModifier.SUSPEND)
                     .addParameter(headerParameterSpec)
@@ -155,7 +155,7 @@ class Generator : CodeGenerator {
                 functions.add(streamingBuilder.build())
             } else if (method.isServerStreaming) {
                 val serverStreamingFunction = FunSpec.builder(method.name.lowerCamelCase())
-                    .addKdoc(sourceInfo.comment())
+                    .addKdoc(sourceInfo.comment().sanitizeKdoc())
                     .addModifiers(KModifier.ABSTRACT)
                     .addModifiers(KModifier.SUSPEND)
                     .addParameter(headerParameterSpec)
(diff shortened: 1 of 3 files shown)


stefanvanburen:
(In draft because I'm still testing this 😅)

timostamm:
The fix looks sensible to me. I'm not sure why whitespace is modified, and I believe `@` should be escaped as well, because it has [meaning in KDoc](https://kotlinlang.org/docs/kotlin-doc.html), but not in proto.

A test for this would be very nice to have. What about a proto file with a service that has offending comments attached? I think it should be generated automatically when put into `protoc-gen-connect-kotlin/proto/buf`, and produce a Kotlin client in `protoc-gen-connect-kotlin/src/test/java`. Then we just need a test case in `protoc-gen-connect-kotlin/src/test/kotlin/PluginGenerationTest.kt` to make sure that the generated code compiles.

stefanvanburen:
ok, added a simple test case in https://github.com/bufbuild/connect-kotlin/pull/39/commits/09cd4bbaee02cdf57a1acd6b7b82d1ad784b926e - happy to extend it in any way that would make us more confident.

stefanvanburen:
generated code looks like this FWIW, which is valid kotlin:

```kotlin
// Code generated by connect-kotlin. DO NOT EDIT.
//
// Source: buf/evilcomments/v1/evilcomments.proto
//
package buf.evilcomments.v1

import build.buf.connect.Headers
import build.buf.connect.MethodSpec
import build.buf.connect.ProtocolClientInterface
import build.buf.connect.ResponseMessage
import build.buf.connect.http.Cancelable
import kotlin.Unit

public class EvilCommentsServiceClient(
  private val client: ProtocolClientInterface,
) : EvilCommentsServiceClientInterface {
  /**
   *  This comment contains characters that should be escaped.
   *  &#64; is valid in KDoc, but not in proto.
   *  Comments in KDoc use C-style block comments, so &#42;/ and /&#42; should be escaped.
   *  \[ and \] characters should also be escaped.
   */
  public override suspend fun evilComments(request: Evilcomments.EvilCommentsRequest,
      headers: Headers): ResponseMessage<Evilcomments.EvilCommentsResponse> = client.unary(
    request,
    headers,
    MethodSpec(
    "buf.evilcomments.v1.EvilCommentsService/EvilComments",
      buf.evilcomments.v1.Evilcomments.EvilCommentsRequest::class,
      buf.evilcomments.v1.Evilcomments.EvilCommentsResponse::class
    ),
  )


  /**
   *  This comment contains characters that should be escaped.
   *  &#64; is valid in KDoc, but not in proto.
   *  Comments in KDoc use C-style block comments, so &#42;/ and /&#42; should be escaped.
   *  \[ and \] characters should also be escaped.
   */
  public override fun evilComments(
    request: Evilcomments.EvilCommentsRequest,
    headers: Headers,
    onResult: (ResponseMessage<Evilcomments.EvilCommentsResponse>) -> Unit,
  ): Cancelable = client.unary(
    request,
    headers,
    MethodSpec(
    "buf.evilcomments.v1.EvilCommentsService/EvilComments",
      buf.evilcomments.v1.Evilcomments.EvilCommentsRequest::class,
      buf.evilcomments.v1.Evilcomments.EvilCommentsResponse::class
    ),
    onResult
  )

}
```

protoc-gen-connect-kotlin/src/main/kotlin/build/buf/protocgen/connect/Generator.kt
```diff
@@ -396,6 +396,18 @@ class Generator : CodeGenerator {
...
+
+    internal fun String.sanitizeKdoc(): String {
+        return this
+            // Remove trailing whitespace on each line.
+            .replace("[^\\S\n]+\n".toRegex(), "\n")
+            .replace("\\s+$".toRegex(), "")
+            .replace("\\*/".toRegex(), "&#42;/")
+            .replace("/\\*".toRegex(), "/&#42;")
+            .replace("""[""", """\[""")
+            .replace("""]""", """\]""")
```
pkwarren, line 408: Are we sure about these two? When I look at the generated source in IDEA, this is how the doc is rendered:

image

pkwarren: I think we may want to do:

```
.replace("""[""", "&#91;")
.replace("""]""", "&#93;")
```

instead.

timostamm (commented):
Should fix the much more pressing issue with `*/` and `/*`, thank you for adding the tests! 

Not sure about escaping `[]` though, see comment below.

pkwarren (approved):

PR #40, opened 2023-04-28
Add extensions for ResponseMessage like Kotlin's Result type

kohenkatz:
The `ResponseMessage` type is very similar to Kotlin's built-in `Result` type. The only  major difference is that the `ResponseMessage` type has additional fields for response code, headers, and trailers.

The Kotlin `Result` type has a number of convenient extension functions which can simplify many common coding situations. For example, `getOrThrow` is useful for situations where a code block is already surrounded by a try/catch, and `getOrNull`/`getOrDefault` are useful for situations in which a failure is not critical. Other places in this library already contain `fold` methods similar to the `Result` type.

This PR adds extension functions to `ResponseMessage` that match the ones present on `Result`.

I'm not sure what type of tests are appropriate for these functions, if any, considering that I don't see any tests for the Kotlin `Result` implementation of these functions.


buildbreaker (approved):
This is a reasonable add. Thanks for your help @kohenkatz

PR #43, opened 2023-05-02
Attempt to fix plugin release to github

pkwarren:
During the last release, we still encountered an issue in the plugin release step. Ensure we set the releaseVersion property when running the gradle build.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,9 +51,11 @@ jobs:
         uses: gradle/wrapper-validation-action@v1
       - name: sha256
         run: |
-          make buildplugin
           tag=$(git describe --tags --abbrev=0 --exact-match)
           release_version="${tag:1}"
+          echo "release_version=${release_version}" >> "$GITHUB_ENV"
+          export ORG_GRADLE_PROJECT_releaseVersion="${release_version}"
+          make buildplugin
           sha256sum ./protoc-gen-connect-kotlin/build/libs/protoc-gen-connect-kotlin-${release_version}.jar >> sha256.txt
       - name: Publish GitHub artifacts
         uses: softprops/action-gh-release@v1


.github/workflows/release.yml
```diff
@@ -51,9 +51,11 @@ jobs:
         uses: gradle/wrapper-validation-action@v1
       - name: sha256
         run: |
-          make buildplugin
           tag=$(git describe --tags --abbrev=0 --exact-match)
           release_version="${tag:1}"
+          echo "release_version=${release_version}" >> "$GITHUB_ENV"
+          export ORG_GRADLE_PROJECT_releaseVersion="${release_version}"
```
pkwarren, line 57: We need to set this prior to running the build to ensure the artifact has the right version number.

.github/workflows/release.yml
```diff
@@ -51,9 +51,11 @@ jobs:
         uses: gradle/wrapper-validation-action@v1
       - name: sha256
         run: |
-          make buildplugin
           tag=$(git describe --tags --abbrev=0 --exact-match)
           release_version="${tag:1}"
+          echo "release_version=${release_version}" >> "$GITHUB_ENV"
```
pkwarren, line 56: `release_version` is used in the next step so we need to set it in the environment.

stefanvanburen: ye, was wondering about that, thanks for the explanation!

pkwarren:
Unfortunately this didn't completely fix the issue - https://github.com/bufbuild/connect-kotlin/pull/44 should resolve it.

I do think however there is merit in improving this in two ways:
1. We should publish the same identical artifact pushed to Maven Central to our GitHub releases.
2. We should update the sha256sum command to be `./sha256sum protoc-gen-connect-kotlin-${release_version}.jar >> sha256.txt` instead of `sha256sum ./protoc-gen-connect-kotlin/build/libs/protoc-gen-connect-kotlin-${release_version}.jar >> sha256.txt`. This will make it easier for people downloading from the artifact to verify the checksum (otherwise they'd have to recreate the directories and move it into there).

pkwarren:
Addressed item 2 in PR #45. Users wanting to download the artifacts in the 0.1.5/0.1.6/0.1.7 releases can download from https://repo1.maven.org/maven2/build/buf/protoc-gen-connect-kotlin/

PR #50, opened 2023-06-02
Add blocking unary generation

buildbreaker:
Providing options for generating synchronous methods for unary RPC calls:
```kotlin
        val response = testServiceConnectClient.unaryCallBlocking(message)
        assertThat(response.code).isEqualTo(Code.UNKNOWN)
        response.failure { errorResponse ->
            assertThat(errorResponse.error).isNotNull()
            assertThat(errorResponse.code).isEqualTo(Code.UNKNOWN)
            assertThat(errorResponse.error.message).isEqualTo("test status message")
        }
        response.success {
            fail<Unit>("unexpected success")
        }
```


rebello95 (approved):
Let's make sure to add docs on this to the Connect site as well

buildbreaker:
@rebello95 Do you prefer "synchronous" or "blocking"?

rebello95:
> @rebello95 Do you prefer "synchronous" or "blocking"?

Don't have a strong preference. I would just avoid `sync`

library/src/main/kotlin/build/buf/connect/Call.kt
```diff
@@ -0,0 +1,41 @@
...
+ * A [Call] contains the way to make a blocking RPC call and cancelling the RPC.
+ */
+class Call<Output> {
+    private var executable: ((ResponseMessage<Output>) -> Unit) -> Unit = { }
+    private var cancel: () -> Unit = { }
+
+    /**
+     * Execute the underlying request.
+     */
+    fun execute(): ResponseMessage<Output> {
```
rebello95, line 16: What prevents this from being called multiple times?

library/src/main/kotlin/build/buf/connect/ProtocolClientInterface.kt
```diff
@@ -56,6 +56,21 @@ interface ProtocolClientInterface {
...
+     * @param headers The outbound request headers to include.
+     * @param methodSpec The Method for RPC path.
+     *
+     * @return The ResponseMessage for the unary call.
+     */
+    fun <Input : Any, Output : Any> unarySync(
+        request: Input,
+        headers: Headers,
+        methodSpec: MethodSpec<Input, Output>
+    ): Call<Output>
```
rebello95, line 72: It feels a little weird to me to have a `Call` class that is only used when a specific (non-default) option is set to `true`, and isn't used in the happy path. If we want something like this, what about making is specific like `BlockingUnaryCall`?

library/src/main/kotlin/build/buf/connect/impl/ProtocolClient.kt
```diff
@@ -117,6 +120,28 @@ class ProtocolClient(
...
     }
 
+    override fun <Input : Any, Output : Any> unarySync(
+        request: Input,
+        headers: Headers,
+        methodSpec: MethodSpec<Input, Output>
+    ): Call<Output> {
+        val countDownLatch = CountDownLatch(1)
+        val reference = AtomicReference<ResponseMessage<Output>>()
+        // Create a new Call object.
```
rebello95, line 130:
```suggestion
```

library/src/main/kotlin/build/buf/connect/impl/ProtocolClient.kt
```diff
@@ -117,6 +120,28 @@ class ProtocolClient(
...
+        val call = Call<Output>()
+        // Set the unary synchronous executable.
+        call.setExecute { callback: (ResponseMessage<Output>) -> Unit ->
+            val cancellationFn = unary(request, headers, methodSpec) { responseMessage ->
```
rebello95, line 134: Not sure if we use `cancellation` or `cancelation` in this repo, but just want to make sure we're consistent everywhere

library/src/main/kotlin/build/buf/connect/impl/ProtocolClient.kt
```diff
@@ -117,6 +120,28 @@ class ProtocolClient(
...
+                callback(responseMessage)
+                reference.set(responseMessage)
+                countDownLatch.countDown()
+            }
+            // Set the cancellation lambda.
```
rebello95, line 139: Might be worth an explanation as to why this is necessary, either here or in the `setCancel` function

library/src/main/kotlin/build/buf/connect/Call.kt
buildbreaker: This isn't prevented.  OkHttp also does this!

elliotmjackson (approved):

README.md
```diff
@@ -106,10 +106,11 @@ is available on the [connect.build website][getting-started].
 
 ## Generation Options 
 
-| **Option**                 | **Type** | **Default** | **Repeatable** | **Details**                                     |
-|----------------------------|:--------:|:-----------:|:--------------:|-------------------------------------------------|
-| `generateCallbackMethods`  | Boolean  |   `false`   |       No       | Generate callback signatures for unary methods. |
-| `generateCoroutineMethods` | Boolean  |   `true`    |       No       | Generate suspend signatures for unary methods.  |
+| **Option**                     | **Type** | **Default** | **Repeatable** | **Details**                                     |
```
rebello95, line 109: Can we make sure this is also documented on the connectrpc website?

crosstests/google-java/src/test/kotlin/build/buf/connect/crosstest/CrossTest.kt
```diff
@@ -376,6 +377,138 @@ class CrossTest(
...
     }
 
+    @Test
+    fun emptyUnaryBlocking(): Unit = runBlocking {
+        val response = testServiceConnectClient.emptyCallBlocking(empty {}).execute()
+        response.failure {
+            fail<Unit>("expected error to be null")
+        }
+        response.success { success ->
+            assertThat(success.message).isEqualTo(empty {})
```
rebello95, line 387: Is there a guarantee that this is actually called? Looks like this test would pass if it was never invoked

rebello95: Same with other tests below

library/src/main/kotlin/build/buf/connect/ProtocolClientInterface.kt
```diff
@@ -56,6 +56,21 @@ interface ProtocolClientInterface {
...
     ): ResponseMessage<Output>
 
+    /**
+     * Perform a synchronous unary (non-streaming) request.
+     *
+     * @param request The outbound request message.
+     * @param headers The outbound request headers to include.
+     * @param methodSpec The Method for RPC path.
+     *
+     * @return The ResponseMessage for the unary call.
```
rebello95, line 66: Incorrect

library/src/main/kotlin/build/buf/connect/UnaryBlockingCall.kt
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * A [UnaryBlockingCall] contains the way to make a blocking RPC call and cancelling the RPC.
+ */
+class UnaryBlockingCall<Output> {
+    private var executable: ((ResponseMessage<Output>) -> Unit) -> Unit = { }
+    private var cancel: () -> Unit = { }
+
+    /**
+     * Execute the underlying request.
```
rebello95, line 28: If this should only be called once, can we document that here?

library/src/main/kotlin/build/buf/connect/impl/ProtocolClient.kt
```diff
@@ -117,6 +120,27 @@ class ProtocolClient(
         }
     }
 
+    override fun <Input : Any, Output : Any> unaryBlocking(
+        request: Input,
+        headers: Headers,
+        methodSpec: MethodSpec<Input, Output>
+    ): UnaryBlockingCall<Output> {
+        val countDownLatch = CountDownLatch(1)
+        val reference = AtomicReference<ResponseMessage<Output>>()
```
rebello95, line 129: Is this necessary?

README.md
```diff
@@ -106,10 +106,11 @@ is available on the [connect.build website][getting-started].
...
+|--------------------------------|:--------:|:-----------:|:--------------:|-------------------------------------------------|
+| `generateCallbackMethods`      | Boolean  |   `false`   |       No       | Generate callback signatures for unary methods. |
```
rebello95, line 111: May as well remove the `Repeatable` column since they're all singular

crosstests/google-java/src/test/kotlin/build/buf/connect/crosstest/CrossTest.kt
buildbreaker: I await for the countdown latch and verify that it got counted down (that the value is zero)

buildbreaker: For the blocking calls, it would call those as they're just for determining if the response was successful or not.

rebello95 (approved):

PR #80, opened 2023-08-23
Add version to user-agent string

pkwarren:


stefanvanburen (approved):

kohenkatz:
If we are changing the User-Agent behavior, can we please allow customization? In other words, can we only set the User-Agent here if it isn't already set, instead of overwriting any user-supplied value in the existing `Headers` object?

stefanvanburen:
seems reasonable to match the [connect-go implementation](https://github.com/connectrpc/connect-go/blob/34b33f6f06d260eaa77104eb3651d8757692418e/protocol_connect.go#L328-L330)!

library/src/main/kotlin/build/buf/connect/protocols/ConnectConstants.kt
```diff
@@ -28,3 +28,10 @@ const val GRPC_STATUS_DETAILS_TRAILERS = "grpc-status-details-bin"
 const val GRPC_STATUS_TRAILER = "grpc-status"
 const val GRPC_TE_HEADER = "te"
 const val GRPC_USER_AGENT = "x-user-agent"
+
+class ConnectConstants private constructor() {
+    companion object {
```
kohenkatz, line 33: Wouldn't this be simpler to write `object ConnectConstraints {` instead of an empty `class` with `private constructor`?

pkwarren:
> If we are changing the User-Agent behavior, can we please allow customization? In other words, can we only set the User-Agent here if it isn't already set, instead of overwriting any user-supplied value in the existing `Headers` object?

Good call - updated and allowed it to be overridden.

library/src/main/kotlin/build/buf/connect/protocols/ConnectConstants.kt
pkwarren: Yup much cleaner - I've written more Java than Kotlin so appreciate you looking out for this sort of thing.

library/src/main/kotlin/build/buf/connect/protocols/GRPCInterceptor.kt
```diff
@@ -188,11 +188,9 @@ internal class GRPCInterceptor(
 
     private fun Headers.withGRPCRequestHeaders(): Headers {
         val headers = toMutableMap()
-        // Note that we do not comply with the recommended structure for user-agent:
-        // https://github.com/grpc/grpc/blob/v1.51.1/doc/PROTOCOL-HTTP2.md#user-agents
-        // But this behavior matches connect-web:
-        // https://github.com/bufbuild/connect-web/blob/v0.4.0/packages/connect-core/src/grpc-web-create-request-header.ts#L33-L36
-        headers[GRPC_USER_AGENT] = listOf("@bufbuild/connect-kotlin")
```
pkwarren, line 195: This header was wrong - gRPC uses the standard `User-Agent` header, only gRPC-web uses a different one.

library/src/main/kotlin/build/buf/connect/protocols/ConnectInterceptor.kt
```diff
@@ -65,6 +65,9 @@ internal class ConnectInterceptor(
                             .map { compressionPool -> compressionPool.name() }
                     )
                 }
+                if (requestHeaders.keys.none { it.equals(USER_AGENT, ignoreCase = true) }) {
```
pkwarren, line 68: Connect requests weren't setting the user agent at all before.

kohenkatz (approved):
I haven't actually tested it yet (though I'm looking forward to doing so), but reading through it looks good to me.

library/src/main/kotlin/build/buf/connect/protocols/GRPCInterceptor.kt
```diff
@@ -188,11 +188,10 @@ internal class GRPCInterceptor(
...
+        if (headers.keys.none { it.equals(USER_AGENT, ignoreCase = true) }) {
+            // https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#user-agents
+            headers[USER_AGENT] = listOf("grpc-kotlin-connect/${ConnectConstants.VERSION}")
```
stefanvanburen, line 193: no strong feeling, but I don't think connect-es/connect-go differentiate the user-agent based on protocol - seems helpful from a debug perspective, just thought I'd call it out as different.

(I think I'd expect maybe `connect-kotlin-grpc`, if we were going to include the protocol, but I guess `grpc-<xxx>` is closer to what the various grpc libraries add?)

library/src/test/kotlin/build/buf/connect/protocols/ConnectInterceptorTest.kt
```diff
@@ -90,6 +90,34 @@ class ConnectInterceptorTest {
         assertThat(request.headers[CONTENT_ENCODING]).isNullOrEmpty()
         assertThat(request.headers["key"]).containsExactly("value")
         assertThat(request.contentType).isEqualTo("content_type")
+        assertThat(request.headers[USER_AGENT]).containsExactly("connect-kotlin/dev")
+    }
+
+    @Test
+    fun requestHeadersCustomUserAgent() {
+        val config = ProtocolClientConfig(
+            host = "https://buf.build",
```
stefanvanburen, line 99: just a note that @smallsamantha will probably want to pick up these as further changes in her sweep in https://github.com/bufbuild/connect-kotlin/pull/79

stefanvanburen (approved):
looks good to me, thanks for adding the tests. feel free to defer to someone with more kotlin experience; two quick comments, neither blocking

library/src/main/kotlin/build/buf/connect/protocols/GRPCInterceptor.kt
pkwarren: Made this change based on connect-go's implementation, alongside the two gRPC specs (which state that it should be `grpc-{language}-{variant}`). I don't feel strongly about this though.

stefanvanburen: ah, yeah, I must have missed it looking through connect-go's source. thanks!

PR #79, opened 2023-08-22
Replace buf.build with connectrpc.com

smallsamantha:
- Update packages build.buf -> com.connectrpc
- Update Eliza usage
- Update buf.build -> connectrpc.com
- Update github references from bufbuild/connect-kotlin to connectrpc/connect-kotlin

PR #84, opened 2023-08-31
Migrate crosstests to use testcontainers

pkwarren:
Instead of manually having to exclude tests from `./gradlew build` (since the crosstest tests fail if the Docker server isn't running), migrate the crosstests to use testcontainers to spin up the server so the tests can always run. This also simplifies the makefile - we no longer need to manually start/stop the crosstests server and also don't need to bind to static ports on the host - it will use ephemeral ports.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,6 @@ MAKEFLAGS += --no-print-directory
 BIN := .tmp/bin
 CACHE := .tmp/cache
 LICENSE_HEADER_YEAR_RANGE := 2022-2023
-CROSSTEST_VERSION := 162d496c009e2ffb1a638b4a2ea789e9cc3331bb
 LICENSE_HEADER_VERSION := v1.26.1
 PROTOC_VERSION ?= 24.1
 GRADLE_ARGS ?=
@@ -26,7 +25,7 @@ $(BIN)/license-headers: Makefile
 
 .PHONY: build
 build: generate ## Build the entire project.
-	./gradlew $(GRADLE_ARGS) build -x test
+	./gradlew $(GRADLE_ARGS) build
 
 .PHONY: buildplugin
 buildplugin: ## Build the connect-kotlin protoc plugin.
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -26,7 +25,7 @@ $(BIN)/license-headers: Makefile
 
 .PHONY: build
 build: generate ## Build the entire project.
-	./gradlew $(GRADLE_ARGS) build -x test
```
pkwarren, line 29: @smallsamantha - I noticed there were real compile errors in the tests which weren't getting reported in any of the CI checks as part of #79. This should resolve that.

Makefile
```diff
@@ -45,19 +44,6 @@ clean: ## Cleans the underlying build.
 
 	rm -rf protoc-gen-connect-kotlin/src/test/java/
 
-.PHONY: crosstestserverrun
-crosstestserverrun: crosstestserverstop ## Run the server for cross tests.
-	docker run --rm --name serverconnect -p 8080:8080 -p 8081:8081 -d \
-		bufbuild/connect-crosstest:$(CROSSTEST_VERSION) \
-		/usr/local/bin/serverconnect --h1port "8080" --h2port "8081" --cert "cert/localhost.crt" --key "cert/localhost.key"
-	docker run --rm --name servergrpc -p 8083:8083 -d \
```
pkwarren, line 53: NOTE: I never saw anything connect to anything other than the h2port of the connect server. If we want to extend the tests to hit the gRPC server we should do that.

crosstests/google-java/src/test/kotlin/build/buf/connect/crosstest/CrossTest.kt
```diff
@@ -71,12 +75,27 @@ class CrossTest(
...
+                "--key",
+                "cert/localhost.key",
+            )
     }
 
     @Before
     fun before() {
-        val port = 8081
-        val host = "https://localhost:$port"
+        val host = "https://localhost:${CROSSTEST_CONTAINER.getMappedPort(8081)}"
```
pkwarren, line 98: This will use an ephemeral port so it won't fail if something else is bound to 8080/8081.

stefanvanburen (approved):

PR #86, opened 2023-09-01
Improve code generation in connect-kotlin

pkwarren:
Generate code to the build/generated/sources directory instead of under src/ to avoid the need for manual cleanup and consolidate where code is placed (so it can be ignored from spotless checks).

Remove generated code from googleapis from source control - we were already generating it with buf.gen.yaml but never configuring the build to use the generated code.


.gitignore
```diff
@@ -3,14 +3,7 @@
...
 # Avoid ignoring Gradle wrapper jar file (.jar files are usually ignored)
 !gradle-wrapper.jar
-/examples/generated-google-java/src/main/
-/examples/generated-google-javalite/src/main/
-/crosstests/google-java/src/main/kotlin/generated/
-/crosstests/google-java/src/main/java/generated/
-/crosstests/google-javalite/src/main/kotlin/generated/
-/crosstests/google-javalite/src/main/java/generated/
 /local.properties
-/protoc-gen-connect-kotlin/src/test/java/
```
pkwarren, line 13: Now that code is generated under `build/` everywhere, it is automatically cleaned up by `./gradlew clean` and we don't need all these other rules in .gitignore.

crosstests/google-javalite/build.gradle.kts
```diff
@@ -0,0 +1,44 @@
+import com.diffplug.gradle.spotless.SpotlessExtension
```
pkwarren, line 1: We were generating java lite code under this directory, but nothing was configured to build it. Added a build file although there may be additional work to do if we want the crosstests to run on the lite code.

stefanvanburen (approved):

PR #87, opened 2023-09-01
Relocation for connect-kotlin

smallsamantha:


smallsamantha:
not going to publish relocation, we'll just move the repository and publish from the new location

Issue #89, opened 2023-09-05
IllegalArgumentException: No property for required constructor parameter


rantianhua:
My gradle config

```
android {
    namespace = "xxx"
    compileSdk = 33

    signingConfigs {
        ...
    }

    defaultConfig {
        applicationId = "xxx"
        minSdk = 26
        targetSdk = 33
        versionCode = 1
        versionName = "1.0.0-debug-v10"

        testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
        vectorDrawables {
            useSupportLibrary = true
        }
        resourceConfigurations.addAll(listOf("en", "ja", "zh-rCN", "zh-rTW"))
    }

    val useMinifyInDebug = hasProperty("minifyInDebug")

    buildTypes {
        debug {
            isMinifyEnabled = useMinifyInDebug
            isShrinkResources = useMinifyInDebug
            isDebuggable = true
            configure<CrashlyticsExtension> {
                mappingFileUploadEnabled = useMinifyInDebug
            }
            signingConfig = signingConfigs.getByName("debug")
            proguardFiles(
                getDefaultProguardFile("proguard-android-optimize.txt"),
                "proguard-rules.pro"
            )
        }
        release {
            isMinifyEnabled = true
            isShrinkResources = true
            configure<CrashlyticsExtension> {
                mappingFileUploadEnabled = true
            }
            signingConfig = signingConfigs.getByName("release")
            proguardFiles(
                getDefaultProguardFile("proguard-android-optimize.txt"),
                "proguard-rules.pro"
            )
        }
    }
    compileOptions {
        sourceCompatibility = JavaVersion.VERSION_1_8
        targetCompatibility = JavaVersion.VERSION_1_8
    }
    kotlinOptions {
        jvmTarget = "1.8"
    }
    kotlin {
        jvmToolchain(8)
    }
    buildFeatures {
        compose = true
        viewBinding = true
        buildConfig = true
    }
    composeOptions {
        // https://developer.android.com/jetpack/androidx/releases/compose-kotlin
        kotlinCompilerExtensionVersion = "1.4.8"
    }
    packaging {
        resources {
            excludes += "/META-INF/{AL2.0,LGPL2.1}"
        }
    }
}

dependencies {
    implementation("build.buf:connect-kotlin-okhttp:0.1.7")
    implementation("build.buf:connect-kotlin-google-javalite-ext:0.1.7")
}
```

[...]

PR #92, opened 2023-09-07
Switch JSON serde to use codegen

pkwarren:
Update JSON serialization/deserialization to use codegen instead of reflection. This improves performance and also generates some proguard rules, so it should help in issue #89.

### DIFF
--- a/library/src/main/kotlin/build/buf/connect/protocols/ConnectInterceptor.kt
+++ b/library/src/main/kotlin/build/buf/connect/protocols/ConnectInterceptor.kt
@@ -46,9 +46,7 @@ import java.net.URL
 internal class ConnectInterceptor(
     private val clientConfig: ProtocolClientConfig
 ) : Interceptor {
-    private val moshi = Moshi.Builder()
-        .add(KotlinJsonAdapterFactory())
-        .build()
+    private val moshi = Moshi.Builder().build()
     private val serializationStrategy = clientConfig.serializationStrategy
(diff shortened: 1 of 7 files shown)


library/src/main/kotlin/build/buf/connect/protocols/ConnectInterceptor.kt
```diff
@@ -46,9 +46,7 @@ import java.net.URL
 internal class ConnectInterceptor(
     private val clientConfig: ProtocolClientConfig
 ) : Interceptor {
-    private val moshi = Moshi.Builder()
-        .add(KotlinJsonAdapterFactory())
```
pkwarren, line 50: Moshi supports either reflection based serialization or serialization using codegen. This removes the factory which powers reflection support.

library/src/main/kotlin/build/buf/connect/protocols/ErrorJSONModels.kt
```diff
@@ -15,18 +15,22 @@
 package build.buf.connect.protocols
 
 import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
 
+@JsonClass(generateAdapter = true)
```
pkwarren, line 20: This annotation will tell the KSP to generate adapter classes (which have toJson/fromJson methods) to support these types.

stefanvanburen (approved):

stefanvanburen (approved):

PR #93, opened 2023-09-08
Fix release build

pkwarren:
Run 'make releaselocal' to verify in CI that a Maven release (with docs) works successfully.


.github/workflows/ci.yml
```diff
@@ -38,6 +38,8 @@ jobs:
           git diff-index --quiet HEAD --
       - name: Build project
         run: make build
+      - name: Install to local repository
+        run: make releaselocal
```
pkwarren, line 42: Added this to CI step so we'll know if the release build is likely to fail. The first commit to this PR only had this change (and you can see it failed the build).

gradle/libs.versions.toml
```diff
@@ -7,7 +7,7 @@ junit = "4.13.2"
 kotlin = "1.6.21"
 kotlinpoet = "1.12.0"
 ktlint = "0.46.0"
-mavenplugin = "0.22.0"
+mavenplugin = "0.24.0"
```
pkwarren, line 10: Bump to a newer plugin. We can't bump to the latest without also bumping to a newer version of the Kotlin plugin (which isn't very straightforward). I'd like to start with this to get a release out.

.github/workflows/ci.yml
rebello95: Can you make the `name` clearer to indicate intent? It's not obvious to me based on the description there now

rebello95 (approved):

PR #95, opened 2023-09-08
Release depends on generated proto code

pkwarren:
Since the googleapis Status files were removed from source control (and are created using 'make generate'), we need to update the release workflow to run generate prior to releasing code.


pkwarren:
Update needed to release workflow after #86.

mfridman (approved):

PR #101, opened 2023-09-15
Fix handling of gRPC trailers-only responses

pkwarren:
connect-kotlin wasn't properly handling gRPC trailers-only responses (which are actually responses with *no* trailers, just headers), leading to the inability to read grpc-status headers properly.

Update the connect-kotlin conformance tests to run combinations of both Connect/gRPC client and server instead of just Connect servers. By enabling this earlier, we would've detected the trailers-only issue earlier with the conformance test.

### DIFF
--- a/conformance/google-java/src/test/kotlin/com/connectrpc/conformance/Conformance.kt
+++ b/conformance/google-java/src/test/kotlin/com/connectrpc/conformance/Conformance.kt
@@ -49,36 +49,38 @@ import org.junit.runner.RunWith
 import org.junit.runners.Parameterized
 import org.junit.runners.Parameterized.Parameters
 import org.testcontainers.containers.GenericContainer
+import org.testcontainers.containers.wait.strategy.HostPortWaitStrategy
 import java.time.Duration
 import java.util.Base64
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
 
-
 @RunWith(Parameterized::class)
 class Conformance(
-    private val protocol: NetworkProtocol
+    private val clientProtocol: NetworkProtocol,
+    private val serverProtocol: NetworkProtocol
 ) {
-    private lateinit var connectClient: ProtocolClient
+    private lateinit var protocolClient: ProtocolClient
     private lateinit var shortTimeoutConnectClient: ProtocolClient
     private lateinit var unimplementedServiceClient: UnimplementedServiceClient
-    private lateinit var testServiceConnectClient: TestServiceClient
+    private lateinit var testServiceClient: TestServiceClient
 
     companion object {
-        const val CONFORMANCE_VERSION = "0b07f579cb61ad89de24524d62f804a2b03b1acf"
+        private const val CONFORMANCE_VERSION = "0b07f579cb61ad89de24524d62f804a2b03b1acf"
 
         @JvmStatic
-        @Parameters(name = "protocol")
-        fun data(): Iterable<NetworkProtocol> {
+        @Parameters(name = "client={0},server={1}")
+        fun data(): Iterable<Array<NetworkProtocol>> {
             return arrayListOf(
-                NetworkProtocol.CONNECT,
-                NetworkProtocol.GRPC
+                arrayOf(NetworkProtocol.CONNECT, NetworkProtocol.CONNECT),
+                arrayOf(NetworkProtocol.GRPC, NetworkProtocol.CONNECT),
+                arrayOf(NetworkProtocol.GRPC, NetworkProtocol.GRPC)
             )
         }
 
         @JvmField
         @ClassRule
-        val CONFORMANCE_CONTAINER = GenericContainer("connectrpc/conformance:$CONFORMANCE_VERSION")
+        val CONFORMANCE_CONTAINER_CONNECT = GenericContainer("connectrpc/conformance:$CONFORMANCE_VERSION")
             .withExposedPorts(8080, 8081)
             .withCommand(
                 "/usr/local/bin/serverconnect",
@@ -91,11 +93,28 @@ class Conformance(
                 "--key",
                 "cert/localhost.key"
             )
+            .waitingFor(HostPortWaitStrategy().forPorts(8081))
+
+        @JvmField
+        @ClassRule
+        val CONFORMANCE_CONTAINER_GRPC = GenericContainer("connectrpc/conformance:$CONFORMANCE_VERSION")
+            .withExposedPorts(8081)
+            .withCommand(
+                "/usr/local/bin/servergrpc",
+                "--port",
+                "8081",
+                "--cert",
+                "cert/localhost.crt",
+                "--key",
+                "cert/localhost.key"
+            )
+            .waitingFor(HostPortWaitStrategy().forPorts(8081))
     }
 
     @Before
     fun before() {
-        val host = "https://localhost:${CONFORMANCE_CONTAINER.getMappedPort(8081)}"
+        val serverPort = if (serverProtocol == NetworkProtocol.CONNECT) CONFORMANCE_CONTAINER_CONNECT.getMappedPort(8081) else CONFORMANCE_CONTAINER_GRPC.getMappedPort(8081)
+        val host = "https://localhost:$serverPort"
         val (sslSocketFactory, trustManager) = sslContext()
         val client = OkHttpClient.Builder()
             .protocols(listOf(Protocol.HTTP_2, Protocol.HTTP_1_1))
@@ -117,23 +136,23 @@ class Conformance(
             ProtocolClientConfig(
                 host = host,
                 serializationStrategy = GoogleJavaProtobufStrategy(),
-                networkProtocol = protocol,
+                networkProtocol = clientProtocol,
                 requestCompression = RequestCompression(10, GzipCompressionPool),
                 compressionPools = listOf(GzipCompressionPool)
             )
         )
-        connectClient = ProtocolClient(
+        protocolClient = ProtocolClient(
             httpClient = ConnectOkHttpClient(client),
             ProtocolClientConfig(
                 host = host,
                 serializationStrategy = GoogleJavaProtobufStrategy(),
-                networkProtocol = protocol,
+                networkProtocol = clientProtocol,
                 requestCompression = RequestCompression(10, GzipCompressionPool),
                 compressionPools = listOf(GzipCompressionPool)
             )
         )
-        testServiceConnectClient = TestServiceClient(connectClient)
-        unimplementedServiceClient = UnimplementedServiceClient(connectClient)
+        testServiceClient = TestServiceClient(protocolClient)
+        unimplementedServiceClient = UnimplementedServiceClient(protocolClient)
     }
 
     @Test
@@ -143,7 +162,7 @@ class Conformance(
             reason = "soirée 🎉"
             domain = "connect-conformance"
         }
-        val stream = testServiceConnectClient.failStreamingOutputCall()
+        val stream = testServiceClient.failStreamingOutputCall()
         val sizes = listOf(
             31415,
             9,
@@ -189,7 +208,7 @@ class Conformance(
     @Test
     fun emptyUnary(): Unit = runBlocking {
         val countDownLatch = CountDownLatch(1)
-        testServiceConnectClient.emptyCall(Empty.newBuilder().build()) { response ->
+        testServiceClient.emptyCall(Empty.newBuilder().build()) { response ->
             response.failure {
                 fail<Unit>("expected error to be null")
             }
@@ -212,7 +231,7 @@ class Conformance(
             }
         }
         val countDownLatch = CountDownLatch(1)
-        testServiceConnectClient.unaryCall(message) { response ->
+        testServiceClient.unaryCall(message) { response ->
             response.failure {
                 fail<Unit>("expected error to be null")
             }
@@ -242,7 +261,7 @@ class Conformance(
             payload = payload { body = ByteString.copyFrom(ByteArray(size)) }
         }
         val countDownLatch = CountDownLatch(1)
-        testServiceConnectClient.unaryCall(message, headers) { response ->
+        testServiceClient.unaryCall(message, headers) { response ->
             assertThat(response.code).isEqualTo(Code.OK)
             assertThat(response.headers[leadingKey]).containsExactly(leadingValue)
             assertThat(response.trailers[trailingKey]).containsExactly(b64Encode(trailingValue))
@@ -267,7 +286,7 @@ class Conformance(
             }
         }
         val countDownLatch = CountDownLatch(1)
-        testServiceConnectClient.unaryCall(message) { response ->
+        testServiceClient.unaryCall(message) { response ->
             assertThat(response.code).isEqualTo(Code.UNKNOWN)
             response.failure { errorResponse ->
                 assertThat(errorResponse.error).isNotNull()
@@ -326,7 +345,7 @@ class Conformance(
         val statusMessage =
             "\\t\\ntest with whitespace\\r\\nand Unicode BMP ☺ and non-BMP \uD83D\uDE08\\t\\n"
         val countDownLatch = CountDownLatch(1)
-        testServiceConnectClient.unaryCall(
+        testServiceClient.unaryCall(
             simpleRequest {
                 responseStatus = echoStatus {
                     code = 2
@@ -352,7 +371,7 @@ class Conformance(
     @Test
     fun unimplementedMethod(): Unit = runBlocking {
         val countDownLatch = CountDownLatch(1)
-        testServiceConnectClient.unimplementedCall(Empty.newBuilder().build()) { response ->
+        testServiceClient.unimplementedCall(Empty.newBuilder().build()) { response ->
             assertThat(response.code).isEqualTo(Code.UNIMPLEMENTED)
             countDownLatch.countDown()
         }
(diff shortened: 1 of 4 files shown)


conformance/google-java/src/test/kotlin/com/connectrpc/conformance/Conformance.kt
```diff
@@ -49,36 +49,38 @@ import org.junit.runner.RunWith
...
-        @Parameters(name = "protocol")
-        fun data(): Iterable<NetworkProtocol> {
+        @Parameters(name = "client={0},server={1}")
+        fun data(): Iterable<Array<NetworkProtocol>> {
             return arrayListOf(
-                NetworkProtocol.CONNECT,
-                NetworkProtocol.GRPC
+                arrayOf(NetworkProtocol.CONNECT, NetworkProtocol.CONNECT),
+                arrayOf(NetworkProtocol.GRPC, NetworkProtocol.CONNECT),
+                arrayOf(NetworkProtocol.GRPC, NetworkProtocol.GRPC)
```
pkwarren, line 77: Now these are the combination of client protocol + server protocol we test. The specific combination which was untested before was a gRPC server.

conformance/google-java/src/test/kotlin/com/connectrpc/conformance/Conformance.kt
```diff
@@ -91,11 +93,28 @@ class Conformance(
                 "--key",
                 "cert/localhost.key"
             )
+            .waitingFor(HostPortWaitStrategy().forPorts(8081))
```
pkwarren, line 96: Added this so we wait long enough for the containers to become healthy.

conformance/google-java/src/test/kotlin/com/connectrpc/conformance/Conformance.kt
```diff
@@ -91,11 +93,28 @@ class Conformance(
...
+
+        @JvmField
+        @ClassRule
+        val CONFORMANCE_CONTAINER_GRPC = GenericContainer("connectrpc/conformance:$CONFORMANCE_VERSION")
```
pkwarren, line 100: Spins up a gRPC conformance server.

library/src/main/kotlin/com/connectrpc/protocols/ConnectInterceptor.kt
```diff
@@ -93,7 +93,7 @@ internal class ConnectInterceptor(
                 trailers.putAll(response.headers.toTrailers())
                 trailers.putAll(response.trailers)
                 val responseHeaders =
-                    response.headers.filter { entry -> !entry.key.startsWith("trailer") }.toMutableMap()
+                    response.headers.filter { entry -> !entry.key.startsWith("trailer") }
```
pkwarren, line 96: Noticed this pattern throughout - we were making headers mutable even though they weren't being mutated.

library/src/main/kotlin/com/connectrpc/protocols/GRPCInterceptor.kt
```diff
@@ -72,30 +72,31 @@ internal class GRPCInterceptor(
                 )
             },
             responseFunction = { response ->
+                val headers = response.headers
                 val trailers = response.trailers
-                val completion = completionParser.parse(trailers)
+                // Handle Trailers-Only response: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses
+                val completion = completionParser.parse(trailers.ifEmpty { headers })
```
pkwarren, line 78: Fix for trailers-only responses.

conformance/google-java/src/test/kotlin/com/connectrpc/conformance/Conformance.kt
```diff
@@ -49,36 +49,38 @@ import org.junit.runner.RunWith
...
 import java.util.Base64
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
 
-
 @RunWith(Parameterized::class)
 class Conformance(
-    private val protocol: NetworkProtocol
+    private val clientProtocol: NetworkProtocol,
+    private val serverProtocol: NetworkProtocol
```
rebello95, line 61: I'm a little confused as to why we need a server and client protocol which differ considering they need to be using the same network protocol. From the looks of the code below, it seems like this is only being used to determine whether to run the server using `connect-go` or `grpc-go`. Would it make sense to run all of these conformance tests for:
- `connect-go`: gRPC, gRPC-Web, Connect protocols
- `grpc-go`: gRPC protocol

For context, this is what Connect-Swift does: https://github.com/connectrpc/connect-swift/blob/fad1776a89fb9e9089ca728c6e55f9fe58f48968/Tests/ConnectLibraryTests/ConnectConformance/ConformanceConfiguration.swift#L46-L59

library/src/main/kotlin/com/connectrpc/protocols/GRPCInterceptor.kt
```diff
@@ -142,7 +143,7 @@ internal class GRPCInterceptor(
             streamResultFunction = { res ->
                 val streamResult = res.fold(
                     onHeaders = { result ->
-                        val responseHeaders = result.headers.filter { entry -> !entry.key.startsWith("trailer") }.toMutableMap()
+                        val responseHeaders = result.headers.filter { entry -> !entry.key.startsWith("trailer") }
```
rebello95, line 146: I think the `onCompletion` parsing is missing the same headers handling as unary now has?

library/src/main/kotlin/com/connectrpc/protocols/GRPCInterceptor.kt
rebello95: @akshayjshah mentioned offline that we should be looking for the `grpc-status` in headers _first_, and then in trailers. Additional context: https://github.com/connectrpc/connect-swift/pull/174

conformance/google-java/src/test/kotlin/com/connectrpc/conformance/Conformance.kt
pkwarren: Yeah was just re-using an existing enum here. This expands from testing:
* client: gRPC, server: Connect
* client: Connect, server: Connect

to also test:
* client: gRPC, server: gRPC

There is follow on work to support gRPC-web and I'd also like to get parameterized tests that run on HTTP/1.1 (for connect) as well.

library/src/main/kotlin/com/connectrpc/protocols/GRPCInterceptor.kt
pkwarren: I'll take a look and alter the algorithm to match.

library/src/main/kotlin/com/connectrpc/protocols/GRPCInterceptor.kt
pkwarren: Ideally this would surface with a failing conformance test. Do you know if there is one that we're perhaps missing here?

rebello95: I believe `failStreamingOutputCall` and `failUnaryCall` fail, but only when the server is `grpc-go`, they pass with `connect-go` since `connect-go` uses trailers

conformance/google-java/src/test/kotlin/com/connectrpc/conformance/Conformance.kt
rebello95: I think it's worth renaming them to indicate that you're trying to hit a `connect-go` or `grpc-go` service because with the existing framing it's convoluting protocols and libraries

pkwarren: Introduced a new ServerType enum.

conformance/google-java/src/test/kotlin/com/connectrpc/conformance/Conformance.kt
```diff
@@ -58,24 +57,25 @@ import java.util.concurrent.TimeUnit
...
         @Parameters(name = "client={0},server={1}")
-        fun data(): Iterable<Array<NetworkProtocol>> {
+        fun data(): Iterable<Array<Any>> {
             return arrayListOf(
-                arrayOf(NetworkProtocol.CONNECT, NetworkProtocol.CONNECT),
-                arrayOf(NetworkProtocol.GRPC, NetworkProtocol.CONNECT),
-                arrayOf(NetworkProtocol.GRPC, NetworkProtocol.GRPC)
+                arrayOf(NetworkProtocol.CONNECT, ServerType.CONNECT_GO),
+                arrayOf(NetworkProtocol.GRPC, ServerType.CONNECT_GO),
+                arrayOf(NetworkProtocol.GRPC_WEB, ServerType.CONNECT_GO),
```
pkwarren, line 77: Added gRPC web protocol tests as well (and found some issues as a result).

conformance/google-java/src/test/kotlin/com/connectrpc/conformance/Conformance.kt
```diff
@@ -383,22 +389,49 @@ class Conformance(
...
+        unimplementedServiceClient.unimplementedCall(empty {}) { response ->
             assertThat(response.code).isEqualTo(Code.UNIMPLEMENTED)
             countDownLatch.countDown()
         }
         countDownLatch.await(500, TimeUnit.MILLISECONDS)
         assertThat(countDownLatch.count).isZero()
     }
 
+    @Test
+    fun unimplementedServerStreamingService(): Unit = runBlocking {
```
pkwarren, line 401: Added this test which caught the streaming behavior issue: https://github.com/connectrpc/connect-kotlin/pull/101#discussion_r1327842292

library/src/main/kotlin/com/connectrpc/protocols/GRPCWebInterceptor.kt
```diff
@@ -276,9 +246,9 @@ internal class GRPCWebInterceptor(
             }
             val i = line.indexOf(":")
             if (i > 0) {
-                val name = line.substring(0, i).trim().lowercase()
-                val value = line.substring(i + 1).trim().lowercase()
```
pkwarren, line 280: This caused issues parsing error details in gRPC-Web (failing conformance tests). We were setting a base64 encoded protobuf to lowercase, which ended up reading as an invalid proto message on the other end.

library/src/main/kotlin/com/connectrpc/protocols/GRPCWebInterceptor.kt
```diff
@@ -259,7 +230,6 @@ internal class GRPCWebInterceptor(
         if (headers.keys.none { it.equals(GRPC_WEB_USER_AGENT, ignoreCase = true) }) {
             headers[GRPC_WEB_USER_AGENT] = listOf("grpc-kotlin-connect/${ConnectConstants.VERSION}")
         }
-        headers[GRPC_TE_HEADER] = listOf("trailers")
```
pkwarren, line 262: gRPC-web specifically doesn't supported trailers - we shouldn't send this header.

library/src/main/kotlin/com/connectrpc/protocols/GRPCWebInterceptor.kt
```diff
@@ -188,26 +185,14 @@ internal class GRPCWebInterceptor(
             streamResultFunction = { res ->
                 val streamResult = res.fold(
                     onHeaders = { result ->
-                        val responseHeaders = result.headers.filter { entry -> !entry.key.startsWith("trailer") }
```
pkwarren: Only Connect, not the gRPC* protocols have this convention with trailers and the `trailer-` prefix.

library/src/main/kotlin/com/connectrpc/protocols/GRPCInterceptor.kt
pkwarren: Updated the algorithm to look at headers first, then trailers.

library/src/main/kotlin/com/connectrpc/protocols/GRPCInterceptor.kt
pkwarren: Added a test which failed because this was missing for streaming calls, and fixed the issue in the streaming implementation.

conformance/google-java/src/test/kotlin/com/connectrpc/conformance/Conformance.kt
```diff
@@ -92,11 +94,28 @@ class Conformance(
...
+                "--key",
+                "cert/localhost.key"
+            )
+            .waitingFor(HostPortWaitStrategy().forPorts(8081))
     }
 
     @Before
     fun before() {
-        val host = "https://localhost:${CONFORMANCE_CONTAINER.getMappedPort(8081)}"
+        val serverPort = if (serverType == ServerType.CONNECT_GO) CONFORMANCE_CONTAINER_CONNECT.getMappedPort(8081) else CONFORMANCE_CONTAINER_GRPC.getMappedPort(8081)
```
rebello95, line 117: Not a big deal, but calling out that Swift uses 8083 for grpc-go because that's where it's hosted in the container: https://github.com/connectrpc/connect-swift/blob/f7aab0e53c38f15d5ab62e8e8afa2f984c7e34f3/Tests/ConnectLibraryTests/ConnectConformance/ConformanceConfiguration.swift#L58

conformance/google-java/src/test/kotlin/com/connectrpc/conformance/Conformance.kt
```diff
@@ -168,14 +187,17 @@ class Conformance(
...
                         onCompletion = { result ->
-                            // For some reason we keep timing out on these calls and not actually getting a real response like with grpc?
-                            assertThat(result.code).isEqualTo(Code.RESOURCE_EXHAUSTED)
-                            assertThat(result.connectError()!!.code).isEqualTo(Code.RESOURCE_EXHAUSTED)
-                            assertThat(result.connectError()!!.message).isEqualTo("soirée 🎉")
-                            assertThat(result.connectError()!!.unpackedDetails(ErrorDetail::class)).containsExactly(
-                                expectedErrorDetail
-                            )
-                            countDownLatch.countDown()
+                            try {
```
rebello95, line 190: Wouldn't we want this to explode if any of the asserts fail rather than allowing tests to continue?

conformance/google-java/src/test/kotlin/com/connectrpc/conformance/Conformance.kt
```diff
@@ -306,10 +328,13 @@ class Conformance(
                 for (res in stream.resultChannel()) {
                     res.maybeFold(
                         onCompletion = { result ->
-                            assertThat(result.error).isNotNull()
-                            assertThat(result.connectError()!!.code).isEqualTo(Code.DEADLINE_EXCEEDED)
-                            assertThat(result.code).isEqualTo(Code.DEADLINE_EXCEEDED)
-                            countDownLatch.countDown()
+                            try {
```
rebello95, line 331: Same here

conformance/google-java/src/test/kotlin/com/connectrpc/conformance/Conformance.kt
```diff
@@ -364,14 +389,41 @@ class Conformance(
...
+    fun unimplementedServerStreamingService(): Unit = runBlocking {
+        val countDownLatch = CountDownLatch(1)
+        val stream = unimplementedServiceClient.unimplementedStreamingOutputCall()
+        stream.send(empty { })
+        withContext(Dispatchers.IO) {
+            val job = async {
+                for (res in stream.resultChannel()) {
+                    res.maybeFold(
+                        onCompletion = { result ->
+                            try {
```
rebello95, line 410: Same question about `try`ing here

library/src/main/kotlin/com/connectrpc/protocols/GRPCCompletion.kt
```diff
@@ -29,5 +32,26 @@ internal data class GRPCCompletion(
...
     val message: ByteString,
     // List of error details.
-    val errorDetails: List<ConnectErrorDetail>
+    val errorDetails: List<ConnectErrorDetail>,
+    // Set to either message headers (or trailers) where the gRPC status was found.
+    val metadata: Headers
 )
+
+internal fun grpcCompletionToConnectError(completion: GRPCCompletion?, serializationStrategy: SerializationStrategy, error: Throwable?): ConnectError? {
+    val code = completion?.code ?: Code.UNKNOWN
```
rebello95, line 41: Could move this down to after the error check since it's not used immediately

library/src/main/kotlin/com/connectrpc/protocols/GRPCCompletionParser.kt
```diff
@@ -27,24 +28,34 @@ class GRPCCompletionParser(
     private val errorDetailParser: ErrorDetailParser
 ) {
     /**
-     * Parses the completion of a GRPC response from the Trailers.
+     * Parses the completion of a GRPC response from the Headers (for trailers-only responses) or Trailers.
      *
      * For GRPCWeb, the caller will have to transform the final message into trailers to parse a completion.
      *
      * For GRPC H2, the caller can just take the trailers from the completed stream to parse a completion.
```
rebello95, line 35: This should probably be updated

rebello95 (approved):

rebello95:
Also, can you please update the PR title/description?

conformance/google-java/src/test/kotlin/com/connectrpc/conformance/Conformance.kt
pkwarren: It will still fail the test, but this will prevent it waiting 5 seconds on the countDownLatch for the failure to register. Verified by adding a failing assert.

rebello95 (approved):

Issue #100, opened 2023-09-15
ClientOnlyStream.receiveAndClose(): Cancellation Exception occurring


Jeremy-Brown14:
I'm getting a crash calling this function, occasionally, it doesn't occur every time but it occurs about ⅓ of the time. It seems that if the function doesn't return the resultChannel.receive() in a timely manner, it hits the finally block and cancels the resultChannel, throwing the below exception. Also, shouldn't this function close the messageStream after receiving the result? I have to call the close() function before calling the receiveAndClose() function, otherwise my client doesn't receive from the result channel and send my backend service does not process the resulting stream I sent to it. Thanks for looking into this!

```
FATAL EXCEPTION: OkHttp Dispatcher
    Process: com.cfa.delivery.driver.qa, PID: 20742
    java.util.concurrent.CancellationException: Channel was cancelled
        at kotlinx.coroutines.channels.BufferedChannel.cancelImpl$kotlinx_coroutines_core(BufferedChannel.kt:1765)
        at kotlinx.coroutines.channels.BufferedChannel.cancel(BufferedChannel.kt:1762)
        at kotlinx.coroutines.channels.ReceiveChannel$DefaultImpls.cancel$default(Channel.kt:297)
        at build.buf.connect.impl.ClientOnlyStream.receiveAndClose(ClientOnlyStream.kt:36)
        at build.buf.connect.impl.ClientOnlyStream$receiveAndClose$1.invokeSuspend(Unknown Source:14)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
        at kotlinx.coroutines.internal.LimitedDispatcher$Worker.run(LimitedDispatcher.kt:115)
        at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:100)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:584)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:793)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:697)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:684)
```

[...]

PR #106, opened 2023-09-20
Fix several streaming issues

pkwarren:
Update conformance tests to add a new test exercising client side streaming, which exposed several issues in streaming call implementations.

The first issue only affected client streaming (it stopped attempting to read a response from the server once the send side closed - it should have stopped only if the receive side closed).

The second issue resulted from not calling close on the channel after the completion message was received, which lead to hangs consuming from `resultChannel()` (it would never complete). After this fix, both examples (for java and javalite) were updated and fixed to correctly exit when finished.

Additionally, several cleanups were made to the API (since the current API for client streaming was non-functional - it would only return the initial Headers result and not the message or completion result).

This should help resolve reported streaming issues like #100.

# API Updates

### Removed
* `close()`
  * Use `sendClose()` instead. This may have confused callers that the close() method would close both send and receive sides of the connection when it was only closing the send side.

### Added
* `sendClose()`
  * This shouldn't typically need to be called as receiveAndClose() already closes the send side of the stream.
* `isSendClosed()`

### Changed
* `receiveAndClose()`
  * Changed to return a ResponseMessage instead of a StreamResult. This allows callers to easily get access to the response as if they were calling a unary method. Previously, the StreamResult would only return the first result retrieved by the call, which typically was a Headers result (leaving callers unable to access the Message or Completion contents). 

### Removed
* `close()`
  * Replaced with `sendClose()`.

### Added
* `receiveClose()`
* `isReceiveClosed()`

### Removed
* `close()`
  * This closed both the send and receive side of the stream (unlike in other interfaces which just closed the send side). If needed, callers should invoke `receiveClose()` instead (although this isn't necessary in normal use).
* `send()`
  * Callers should invoke `sendAndClose()` instead. Otherwise, reading results from `resultChannel()` will hang since the send side of the stream should be closed before reading responses.

### Removed
* Removed the `error` field from the base `StreamResult` class. It was never used by the `Headers` or `Message` subclasses and only used on the `Complete` type. This should make it easier for callers to use `Headers` and `Message`  types since they don't need to worry about handling `error`.


library/src/main/kotlin/com/connectrpc/impl/ProtocolClient.kt
```diff
@@ -211,6 +211,9 @@ class ProtocolClient(
                 }
             }
             channel.send(result)
+            if (isComplete) {
+                channel.close()
```
pkwarren, line 215: This was a major issue - not closing the channel would lead to hangs in streaming clients trying to read results.

okhttp/src/main/kotlin/com/connectrpc/okhttp/OkHttpStream.kt
```diff
@@ -61,7 +61,7 @@ internal fun OkHttpClient.initializeStream(
     }
     val callRequest = builder.build()
     val call = newCall(callRequest)
-    call.enqueue(ResponseCallback(onResult, isSendClosed))
+    call.enqueue(ResponseCallback(onResult, isReceiveClosed))
```
pkwarren, line 64: This was the other big error for client streaming calls - this was using `isSendClosed` instead of `isReceiveClosed`, so we would stop trying to read messages from a stream once the send side was closed.

conformance/google-java/src/test/kotlin/com/connectrpc/conformance/Conformance.kt
```diff
@@ -754,6 +750,87 @@ class Conformance(
...
+                    countDownLatch.countDown()
+                }
+            }
+            countDownLatch.await(5, TimeUnit.MINUTES)
+            job.cancel()
+            assertThat(countDownLatch.count).isZero()
+        }
+    }
+
+    data class ServerStreamingResult<Output>(
```
rebello95, line 784:
```suggestion
    private data class ServerStreamingResult<Output>(
```

conformance/google-java/src/test/kotlin/com/connectrpc/conformance/Conformance.kt
```diff
@@ -754,6 +750,87 @@ class Conformance(
...
+     * manually in each location.
+     */
+    private suspend fun <Output> streamResults(channel: ReceiveChannel<StreamResult<Output>>): ServerStreamingResult<Output> {
+        val seenHeaders = AtomicBoolean(false)
+        var headers: Headers = emptyMap()
+        val messages: MutableList<Output> = mutableListOf()
+        val seenCompletion = AtomicBoolean(false)
+        var code: Code = Code.UNKNOWN
+        var trailers: Headers = emptyMap()
+        var error: Throwable? = null
```
rebello95, line 806:
```suggestion
        var error: Throwable?
```

examples/kotlin-google-java/src/main/kotlin/com/connectrpc/examples/kotlin/Main.kt
```diff
@@ -33,37 +35,50 @@ class Main {
...
                             val elizaResponse = result.message
                             println(elizaResponse.sentence)
                         },
+                        onCompletion = { result ->
+                            if (result.code != Code.OK) {
+                                val connectErr = result.connectError()
+                                if (connectErr != null) {
+                                    throw connectErr
+                                }
+                                throw ConnectError(code = result.code, metadata = result.trailers)
```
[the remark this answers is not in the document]
rebello95: Maybe just print the error details as a response?

rebello95 (approved):

examples/kotlin-google-java/src/main/kotlin/com/connectrpc/examples/kotlin/Main.kt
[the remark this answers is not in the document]
pkwarren: I wanted to illustrate how you might want to handle errors in real code (instead of printing code/cause/trailers).

I think it also points out opportunities to make it even easier for consumers - if the Code is ever not OK, we should always have a ConnectError (even if it wraps another cause).

rebello95 (approved):

PR #108, opened 2023-09-22
Add additional conformance streaming tests

pkwarren:
Add a server streaming successful test case and a ping/pong test case for verifying bidirectional streams.


conformance/google-java/src/test/kotlin/com/connectrpc/conformance/Conformance.kt
```diff
@@ -775,7 +806,7 @@ class Conformance(
                     countDownLatch.countDown()
                 }
             }
-            countDownLatch.await(5, TimeUnit.MINUTES)
```
pkwarren, line 778: Had set this higher while debugging - can revert back to more reasonable wait times.

rebello95 (approved):

PR #120, opened 2023-09-28
Rework ConnectError to ConnectException

jzbrooks:
Exceptions that might be caught should not extend throwable. Throwable
is the subtype of both Exception and Error. The latter indicates a
critical failure and should only be caught in very rare circumstances.
This change renames the class and makes the supertype more specific
to suggest that catching this throwable is reasonable.

**Before submitting your PR:** Please read through the contribution guide at https://github.com/connectrpc/connect-kotlin/blob/main/.github/CONTRIBUTING.md

**Note:** This is a source-breaking change, which was briefly discussed in slack.

### DIFF
--- a/library/src/main/kotlin/com/connectrpc/StreamResult.kt
+++ b/library/src/main/kotlin/com/connectrpc/StreamResult.kt
@@ -37,7 +37,7 @@ sealed class StreamResult<Output> {
     // Stream is complete. Provides the end status code and optionally an error and trailers.
     class Complete<Output>(val code: Code, val error: Throwable? = null, val trailers: Trailers = emptyMap()) : StreamResult<Output>() {
         /**
-         * Get the ConnectError from the result.
+         * Get the ConnectException from the result.
          *
          * @return The [ConnectException] if present, null otherwise.
          */
--- a/library/src/main/kotlin/com/connectrpc/StreamResult.kt
+++ b/library/src/main/kotlin/com/connectrpc/StreamResult.kt
@@ -39,10 +39,10 @@ sealed class StreamResult<Output> {
         /**
          * Get the ConnectError from the result.
          *
-         * @return The [ConnectError] if present, null otherwise.
+         * @return The [ConnectException] if present, null otherwise.
          */
-        fun connectError(): ConnectError? {
-            if (error is ConnectError) {
+        fun connectError(): ConnectException? {
+            if (error is ConnectException) {
                 return error
             }
             return null
--- a/library/src/main/kotlin/com/connectrpc/StreamResult.kt
+++ b/library/src/main/kotlin/com/connectrpc/StreamResult.kt
@@ -35,16 +35,16 @@ sealed class StreamResult<Output> {
     }
 
     // Stream is complete. Provides the end status code and optionally an error and trailers.
-    class Complete<Output>(val code: Code, val error: Throwable? = null, val trailers: Trailers = emptyMap()) : StreamResult<Output>() {
+    class Complete<Output>(val code: Code, val cause: Throwable? = null, val trailers: Trailers = emptyMap()) : StreamResult<Output>() {
         /**
          * Get the ConnectException from the result.
          *
          * @return The [ConnectException] if present, null otherwise.
          */
-        fun connectException() = error as? ConnectException
+        fun connectException() = cause as? ConnectException
 
         override fun toString(): String {
-            return "Complete{code=$code,error=$error,trailers=$trailers}"
+            return "Complete{code=$code,cause=$cause,trailers=$trailers}"
         }
     }
 
--- a/library/src/main/kotlin/com/connectrpc/StreamResult.kt
+++ b/library/src/main/kotlin/com/connectrpc/StreamResult.kt
@@ -41,12 +41,7 @@ sealed class StreamResult<Output> {
          *
          * @return The [ConnectException] if present, null otherwise.
          */
-        fun connectError(): ConnectException? {
-            if (error is ConnectException) {
-                return error
-            }
-            return null
-        }
+        fun connectException() = error as? ConnectException
 
         override fun toString(): String {
             return "Complete{code=$code,error=$error,trailers=$trailers}"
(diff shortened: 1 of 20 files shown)


library/src/main/kotlin/com/connectrpc/StreamResult.kt
```diff
@@ -39,10 +39,10 @@ sealed class StreamResult<Output> {
         /**
          * Get the ConnectError from the result.
```
pkwarren, line 40:
```suggestion
         * Get the ConnectException from the result.
```

pkwarren (commented):
There are several other APIs that we should consider updating to use consistent names everywhere - i.e. `connectError()` -> `connectException()`.

Since we just made an API change in the last release, I wonder if we shouldn't hold off on changing `ConnectError` -> `ConnectException` everywhere so we can accumulate other API changes as well prior to stabilizing the API for a v1. Maybe we split this PR into two - first doing the Throwable -> Exception change and then saving another with a full rename to merge along with other API changes.

jzbrooks:
Changing the super type isn't a source breaking change, but it does break the API contract because some people might be depending on the super type being `Throwable`. If we make that breaking change, in my opinion we might as well make the source breaking change to get them all out of the way quickly.

I'll look through a little more carefully for other parts of the API surface that need a name update.

I get the concern with a breaking change (especially successive ones). I understand if we want to leave this open a while. Have y'all considered keeping a changelog in the repo? It might make changes like this one more palatable for early adopters.

pkwarren (commented):
A few more spots need to be updated:
* `grpcCompletionToConnectError` -> `grpcCompletionToConnectException`.
* `connectErr` -> `connectExcception` in conformance tests and the Main.kt examples.
* `ServerStreamingResult.error` -> `exception` in the conformance tests.
* `parseConnectUnaryError` -> `parseConnectUnaryException`.

I think we should also change the following for consistency:
* `ResponseMessage.Failure.error` -> `exception` or `cause`.
* `StreamResult.Complete.error` -> `exception` or `cause`.
* `HTTPResponse.error` -> `exception` or `cause`.

jzbrooks:
I think the remaining issues are addressed. I was able to verify the conformance tests compile, but I'm not able to run them. I'll keep an eye on the PR checks when they're run. Thanks!

Note the extended commit message in 4c10702dbf3a93d5b339901d803341a0af79110c

library/src/main/kotlin/com/connectrpc/protocols/GRPCCompletion.kt
```diff
@@ -35,23 +35,27 @@ internal data class GRPCCompletion(
     val errorDetails: List<ConnectErrorDetail>,
     // Set to either message headers (or trailers) where the gRPC status was found.
     val metadata: Headers,
-)
+) {
+    /**
+     * Converts a completion into a [ConnectException] if the completion failed or if a throwable is passed
+     * @return a ConnectException on failure, null otherwise
+     */
+    fun toConnectExceptionOrNull(serializationStrategy: SerializationStrategy, cause: Throwable? = null): ConnectException? {
```
pkwarren, line 43: Nice refactoring here.

pkwarren (approved):
Looks great!

pkwarren:
Looks like the lint check is failing - if you run `make lintfix` and push the changes that should fix it.

pkwarren (changes requested):
Blocking for now until we can address conformance tests and lint warnings.

jzbrooks:
The assertions in the conformance tests don't print the assertion message. This makes debugging via CI difficult. I don't have a docker license, so I can't run them locally.

library/src/main/kotlin/com/connectrpc/protocols/GRPCInterceptor.kt
```diff
@@ -161,12 +161,20 @@ internal class GRPCInterceptor(
...
+                                serializationStrategy,
+                                result.cause,
+                            )
+                            StreamResult.Complete(
+                                code = exception?.code ?: Code.OK,
+                                cause = exception,
+                                trailers = trailers,
+                            )
+                        } else {
+                            val exception = ConnectException(Code.UNKNOWN)
```
pkwarren, line 175: To resolve the test failure, this else condition should change to:

```kotlin
return@fold result
```

This way we'll preserve errors like `DEADLINE_EXCEEDED`.

library/src/main/kotlin/com/connectrpc/protocols/GRPCInterceptor.kt
```diff
@@ -161,15 +163,22 @@ internal class GRPCInterceptor(
...
+                                result.cause,
+                            )
+                            StreamResult.Complete(
+                                code = exception?.code ?: Code.OK,
+                                cause = exception,
+                                trailers = trailers,
+                            )
+                        } else {
+                            val exception = ConnectException(result.code)
+                            StreamResult.Complete(exception.code, exception, trailers)
```
pkwarren, line 178: We may be losing an underlying cause here - so we should just return the result here.

library/src/main/kotlin/com/connectrpc/protocols/GRPCInterceptor.kt
pkwarren: So result is already instantiated in some cases (with a code like DEADLINE_EXCEEDED and a cause). Some of the tests rely on this. Alternatively to doing `return@fold`. you could just have `result` here as the return value.

library/src/main/kotlin/com/connectrpc/protocols/GRPCInterceptor.kt
jzbrooks: I worked around the problem by using the result's code property value in the connect exception constructor, but I like that idea better.

e83429aa067fa83cda356955328eb0d0f96bc0ec

pkwarren (approved):

library/src/main/kotlin/com/connectrpc/protocols/GRPCInterceptor.kt
jzbrooks: I'm tracking now. Thanks for the explaination.

PR #128, opened 2023-10-10
Allow separate client for streaming calls

### LINKED ISSUE #13: Allow enabling HTTP/2 PING and disabling read timeout for streams
The default okhttp `readTimeout` is 10 seconds, which is too short for many uses of streaming RPCs. As an alternative way to detect broken connections, HTTP/2 provides a `PING` mechanism. okhttp has pings disabled by default, but they can be enabled by calling `pingInterval()` on the builder. Streaming calls can set `readTimeout` to `0` to disable the read timeout and rely only on the ping timeout, if desired.

The problem is that there is not an easy way to change the okhttp settings for unary vs. streaming calls, and we still want to have a read timeout for the unary calls.

I thought it would be easiest to fix this by extending `ConnectOkHttpClient` to customize the `client` object and then call the parent method using the customized client, but `ConnectOkHttpClient` would need to be `open` instead of `final` to allow this.

I ended up doing this, which feels like a nasty hack:

```kotlin
class PingingConnectClient(client: OkHttpClient): HTTPClientInterface {
    private val internalUnaryClient = ConnectOkHttpClient(client)
    private val internalStreamClient = ConnectOkHttpClient(client.newBuilder()
        .pingInterval(30, TimeUnit.SECONDS)
        .readTimeout(0, TimeUnit.SECONDS)
        .build())

    override fun unary(request: HTTPRequest, onResult: (HTTPResponse) -> Unit): Cancelable {
        return internalUnaryClient.unary(request, onResult)
    }

    override fun stream(
        request: HTTPRequest,
        onResult: suspend (StreamResult<Buffer>) -> Unit
    ): Stream {
        return internalStreamClient.stream(request, onResult)
    }
}
```

pkwarren:
Streaming calls often require longer timeouts (for long lived streams) and enabling pings (to keep the underlying connection alive). Update ConnectOkHttpClient to take a second OkHttpClient argument to the constructor to be used for streaming calls.

Fixes #13.

### DIFF
--- a/okhttp/src/main/kotlin/com/connectrpc/okhttp/ConnectOkHttpClient.kt
+++ b/okhttp/src/main/kotlin/com/connectrpc/okhttp/ConnectOkHttpClient.kt
@@ -38,7 +38,8 @@ import java.io.IOException
  * The OkHttp implementation of HTTPClientInterface.
  */
 class ConnectOkHttpClient @JvmOverloads constructor(
-    val client: OkHttpClient = OkHttpClient(),
+    private val unaryClient: OkHttpClient = OkHttpClient(),
+    private val streamClient: OkHttpClient = unaryClient,
 ) : HTTPClientInterface {
 
     override fun unary(request: HTTPRequest, onResult: (HTTPResponse) -> Unit): Cancelable {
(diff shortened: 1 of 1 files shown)


rebello95 (commented):
Is it worth updating conformance tests to specify a different client for streaming?

pkwarren:
> Is it worth updating conformance tests to specify a different client for streaming?

Without some sort of long running test that relies on ping to keep it alive, I don't know if it is worth it.

rebello95 (approved):

PR #130, opened 2023-10-18
Improve API for bidi and server streaming calls

pkwarren:
Instead of requiring callers to handle oneOf(Headers,Message,Trailers) objects in each bidi or server streaming call, instead just change the response channel to return the response message type. If an error occurs at the end of the call (due to non-zero grpc-status), then cancel the channel with an exception.

### DIFF
--- a/library/src/main/kotlin/com/connectrpc/BidirectionalStreamInterface.kt
+++ b/library/src/main/kotlin/com/connectrpc/BidirectionalStreamInterface.kt
@@ -21,11 +21,11 @@ import kotlinx.coroutines.channels.ReceiveChannel
  */
 interface BidirectionalStreamInterface<Input, Output> {
     /**
-     * The Channel for received StreamResults.
+     * The Channel for responses.
      *
-     * @return ReceiveChannel for iterating over the received results.
+     * @return ReceiveChannel for iterating over the responses.
      */
-    fun resultChannel(): ReceiveChannel<StreamResult<Output>>
+    fun responseChannel(): ReceiveChannel<Output>
 
     /**
      * Send a request to the server over the stream.
--- a/library/src/main/kotlin/com/connectrpc/impl/ClientOnlyStream.kt
+++ b/library/src/main/kotlin/com/connectrpc/impl/ClientOnlyStream.kt
@@ -34,7 +34,7 @@ internal class ClientOnlyStream<Input, Output>(
         try {
             messageStream.sendClose()
             val message = resultChannel.receive()
-            val additionalMessage = resultChannel.tryReceive()
+            val additionalMessage = resultChannel.receiveCatching()
             if (additionalMessage.isSuccess) {
                 throw ConnectException(code = Code.UNKNOWN, message = "unary stream has multiple messages")
             }
--- a/library/src/main/kotlin/com/connectrpc/impl/ClientOnlyStream.kt
+++ b/library/src/main/kotlin/com/connectrpc/impl/ClientOnlyStream.kt
@@ -18,8 +18,6 @@ import com.connectrpc.BidirectionalStreamInterface
 import com.connectrpc.ClientOnlyStreamInterface
 import com.connectrpc.Code
 import com.connectrpc.ConnectException
-import com.connectrpc.Headers
-import com.connectrpc.ResponseMessage
 
 /**
  * Concrete implementation of [ClientOnlyStreamInterface].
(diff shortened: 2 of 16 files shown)


library/src/main/kotlin/com/connectrpc/impl/ClientOnlyStream.kt
```diff
@@ -31,59 +29,16 @@ internal class ClientOnlyStream<Input, Output>(
...
-                            } else {
-                                ConnectException(code ?: Code.UNKNOWN, message = error?.message, exception = error, metadata = trailers)
-                            }
-                        }
-                    },
-                )
+            val message = resultChannel.receive()
+            val additionalMessage = resultChannel.receiveCatching()
+            if (additionalMessage.isSuccess) {
+                throw ConnectException(code = Code.UNKNOWN, message = "unary stream has multiple messages")
```
jzbrooks, line 39: It seems strange to that we'd suggest the reason is unknown (per the code) when we know what's going on here.

pkwarren: This matches the behavior from connect-go: https://github.com/connectrpc/connect-go/blob/96effedc8ac84da9f49392e5f7bdfee2e648247b/connect.go#L367C1-L368

I think in this case, someone has broken the contract of what a client streaming call should adhere to, so we return UNKNOWN. I don't think any other codes make sense for us to return here, as they're typically used by server implementations.

library/src/main/kotlin/com/connectrpc/impl/ClientOnlyStream.kt
```diff
@@ -31,59 +29,16 @@ internal class ClientOnlyStream<Input, Output>(
...
-                            error = if (resultErr is ConnectException) {
-                                resultErr
-                            } else {
-                                ConnectException(code ?: Code.UNKNOWN, message = error?.message, exception = error, metadata = trailers)
-                            }
-                        }
-                    },
-                )
+            val message = resultChannel.receive()
+            val additionalMessage = resultChannel.receiveCatching()
```
rebello95, line 37: Not sure I understand why we have `message` and `additionalMessage` here - can you elaborate on this?

library/src/main/kotlin/com/connectrpc/BidirectionalStreamInterface.kt
```diff
@@ -21,11 +21,11 @@ import kotlinx.coroutines.channels.ReceiveChannel
...
 interface BidirectionalStreamInterface<Input, Output> {
     /**
-     * The Channel for received StreamResults.
+     * The Channel for responses.
      *
-     * @return ReceiveChannel for iterating over the received results.
+     * @return ReceiveChannel for iterating over the responses.
      */
-    fun resultChannel(): ReceiveChannel<StreamResult<Output>>
+    fun responseChannel(): ReceiveChannel<Output>
```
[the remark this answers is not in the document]
pkwarren: Consumers can continue to access stream response headers from interceptors - I've opened https://github.com/connectrpc/connect-kotlin/issues/131 with some ideas on making it easier for people to access these based on patterns in connect-go.

library/src/main/kotlin/com/connectrpc/impl/ClientOnlyStream.kt
pkwarren: For client only streaming, the server should send only a single response. This just ensures that a server implementation doesn't send > 1 response (this validation was missing before).

rebello95 (approved):
I think overall these changes are fine, but definitely think #131 should be prioritized after merging this.

Can we delete `StreamResult` now?

I also want to call out that Swift provides a similar `StreamResult` [interface](https://github.com/connectrpc/connect-swift/blob/2d072c69bd89a3dd9274574e6a5bdcac873a8514/Libraries/Connect/Interfaces/Streams/StreamResult.swift#L19), but I think it's easier for clients to consume with Swift's APIs. To demonstrate: https://github.com/connectrpc/connect-swift/pull/202

rebello95:
Would also like to see if @jzbrooks has any feedback on this interface

kohenkatz (commented):
Thanks to @pkwarren for tagging me on Slack too look at this. I like the simplification here, and it will definitely help some of my coworkers who were having trouble understanding our app code for bidi streaming.

pkwarren:
> Can we delete `StreamResult` now?

Not at the moment - it is still used by the interceptor API.

PR #139, opened 2023-10-31
Support empty buffers on Gzip decompression

### LINKED ISSUE #138: GzipCompressionPool decompress fails with EOFException on empty Buffer
When gzip compression is enabled and compressed, empty responses are returned, `ConnectInterceptor`s response parsing fails due to `GzipCompressionPool.decompress` not being able to handle empty Buffers.

It's unclear to me if this is actually a valid response message state that connect-kotlin should handle, or if this is just a symptom of our QA environment being in a weird state serving back invalid responses.

Feel free to close if this is invalid

erawhctim:
Fixes #138 

Gzipped responses that return an empty message Buffer fail with an `EOFException`, halting the request chain. 

The fix is relatively simple, but I'm not sure if this is actually a valid state connect-kotlin should support


okhttp/src/test/kotlin/com/connectrpc/okhttp/MockWebServerTests.kt
```diff
@@ -0,0 +1,76 @@
...
+    fun `compressed empty failure response is parsed correctly`() = runTest {
+        val mockWebServer = MockWebServer()
+        mockWebServer.start()
+
+        mockWebServer.enqueue(
+            MockResponse().apply {
+                addHeader("content-encoding", "gzip")
+                addHeader("content-type", "application/json")
+                setBody(Buffer())
+                setResponseCode(401)
```
erawhctim, line 46: This isn't quite representative of the original scenario we were seeing: the server was returning a non-empty response, but by the time it got [here](https://github.com/connectrpc/connect-kotlin/blob/7b7c746352c83bed44cb4659f6ba5566750c3e5b/library/src/main/kotlin/com/connectrpc/protocols/ConnectInterceptor.kt#L104), the response message buffer was already empty. 

Our hypothesis after stepping through the code a few times, was that `parseConnectUnaryException()` was consuming the message buffer before the decompress call had a change to consume it: the message was always non-empty when `parseConnectUnaryException` was called. We thought the root cause might be to decompress the response _first_, then attempt to parse it into an error message. I haven't been able to reproduce that exact scenario in this test, though; the _actual_ root cause may still be TBD.

okhttp/build.gradle.kts
```diff
@@ -16,6 +16,12 @@ dependencies {
     implementation(libs.kotlin.coroutines.core)
 
     api(project(":library"))
+
+    testImplementation(libs.assertj)
+    testImplementation(libs.okhttp.mockwebserver)
+    testImplementation(libs.kotlin.coroutines.test)
+    testImplementation(project(":extensions:google-java"))
+    testImplementation(project(":conformance:google-java"))
```
pkwarren, line 24: I don't think we want the okhttp module to depend on the conformance module - perhaps we just need a simple Eliza codegen in okhttp for tests so we have a service we can use.

okhttp/src/test/kotlin/com/connectrpc/okhttp/MockWebServerTests.kt
```diff
@@ -0,0 +1,76 @@
...
+import okhttp3.mockwebserver.MockWebServer
+import okio.Buffer
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+
+class MockWebServerTests {
+
+    @Test
+    fun `compressed empty failure response is parsed correctly`() = runTest {
+        val mockWebServer = MockWebServer()
```
pkwarren, line 38: We should wire up https://github.com/square/okhttp/tree/master/mockwebserver-junit4 (or even better migrate these tests to junit jupiter and use https://github.com/square/okhttp/tree/master/mockwebserver-junit5. This will make it easier over time to add more tests here without worrying about construction/start/shutdown being called in each.

okhttp/src/test/kotlin/com/connectrpc/okhttp/MockWebServerTests.kt
```diff
@@ -0,0 +1,76 @@
...
+                host = host.toString(),
+                serializationStrategy = GoogleJavaProtobufStrategy(),
+                networkProtocol = NetworkProtocol.CONNECT,
+                requestCompression = RequestCompression(0, GzipCompressionPool),
+                compressionPools = listOf(GzipCompressionPool),
+            ),
+        )
+
+        val request = simpleRequest {}
+        TestServiceClient(protocolClient).unaryCall(request)
```
pkwarren, line 68: Don't we want to assert something here about the returned call (that it has an appropriate connect status code)?

okhttp/build.gradle.kts
erawhctim: Sure, that makes sense. Would that be set up similarly to how the `:examples:generated-google-java` and `:examples:generated-google-javalite` modules are configured?

okhttp/src/test/kotlin/com/connectrpc/okhttp/MockWebServerTests.kt
erawhctim: Great idea. Do you have a preference over JUnit 4 vs. 5?

okhttp/src/test/kotlin/com/connectrpc/okhttp/MockWebServerTests.kt
erawhctim: Yep, I can add some more detailed assertions here

okhttp/src/test/kotlin/com/connectrpc/okhttp/MockWebServerTests.kt
erawhctim: Ended up using JUnit4. I think it makes sense to migrate everything over to JUnit Jupiter in a future PR.

gradle/libs.versions.toml
```diff
@@ -9,6 +9,7 @@ kotlinpoet = "1.14.2"
 mavenplugin = "0.25.3"
 moshi = "1.15.0"
 okhttp = "4.12.0"
+okhttp-junit = "5.0.0-alpha.11"
```
erawhctim, line 12: lemme know if you'd rather keep the okhttp versions the same (and depend on the older `mockwebserver` artifact)

pkwarren: Yeah let's keep them in sync.

okhttp/src/test/kotlin/com/connectrpc/okhttp/MockWebServerTests.kt
```diff
@@ -0,0 +1,74 @@
...
+            ),
+        )
+
+        val response = ElizaServiceClient(protocolClient).say(sayRequest { sentence = "hello" })
+
+        mockWebServerRule.server.takeRequest().apply {
+            assertThat(path).isEqualTo("/connectrpc.eliza.v1.ElizaService/Say")
+        }
+
+        assertThat(response.code).isEqualTo(Code.UNKNOWN)
```
pkwarren, line 72: Wrote a small connect-go example and verified that it gets the same result here. Thanks for adding these tests - they'll make it much easier to test different edge conditions outside of the conformance tests in the future.

pkwarren (approved):

PR #145, opened 2023-11-06
Fix conformance test and update to protoc 25.0

pkwarren:


conformance/common/src/main/kotlin/com/connectrpc/conformance/ssl/SSL.kt
```diff
@@ -55,96 +58,3 @@ fun sslContext(): Pair<SSLSocketFactory, X509TrustManager> {
 
     return result.sslSocketFactory() to result.trustManager
 }
-
-// https://github.com/connectrpc/conformance/blob/main/cert/client.crt
-// cert issues: https://stackoverflow.com/questions/9210514/unable-to-find-valid-certification-path-to-requested-target-error-even-after-c
-private const val CLIENT_CERT = """-----BEGIN CERTIFICATE-----
```
pkwarren, line 61: Changed to just read these in as resources instead of including them as string constants here. This will make it easier to update if we ever need to do so again.

conformance/google-java/src/test/kotlin/com/connectrpc/conformance/Conformance.kt
```diff
@@ -72,7 +72,7 @@ class Conformance(
     private lateinit var testServiceConnectClient: TestServiceClient
 
     companion object {
-        private const val CONFORMANCE_VERSION = "0b07f579cb61ad89de24524d62f804a2b03b1acf"
+        private const val CONFORMANCE_VERSION = "88f85130640b46c0837e0d58c0484d83a110f418"
```
pkwarren, line 75: New upstream version with updated certs (will expire in 10 years).

conformance/common/src/main/kotlin/com/connectrpc/conformance/ssl/SSL.kt
smaye81: Good call.

smaye81 (approved):

PR #146, opened 2023-11-08
Allow building with Java 20 and later versions

pkwarren:
Update the KotlinCompile args to add `-Xjdk-release=1.8` which is the equivalent of the `--release 8` argument when compiling Java code. This will ensure that we only use APIs which are in Java 1.8 and not added in later versions. This fixes compiler warnings around use of `new URL(String)` which is deprecated starting in Java 20.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -133,6 +133,9 @@ subprojects {
             jvmTarget = "1.8"
             languageVersion = "1.6"
             apiVersion = "1.6"
+            if (JavaVersion.current().isJava9Compatible && project.name != "android") {
+                freeCompilerArgs += "-Xjdk-release=1.8"
+            }
         }
     }
     tasks.withType<JavaCompile> {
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -133,6 +133,9 @@ subprojects {
             jvmTarget = "1.8"
             languageVersion = "1.6"
             apiVersion = "1.6"
+            if (JavaVersion.current().isJava9Compatible && project.name != "android") {
```
pkwarren, line 136: Enabling this for android resulted in obscure errors:
```
> Task :examples:android:compileDebugKotlin FAILED
e: JDK_HOME path is not specified in compiler configuration
```

I'm guessing that it may use a different compiler version that doesn't support this argument.

erawhctim: Is this something that's resolved (or simplified) with Gradle's [JVM Toolchain support](https://kotlinlang.org/docs/gradle-configure-project.html#gradle-java-toolchains-support)?

erawhctim (approved):

build.gradle.kts
pkwarren: I find that the toolchain support does more than I want in local testing - it is a bit too prescriptive.

smaye81 (approved):

PR #172, opened 2023-12-07
Add method stream type to generated code

pkwarren:
It is useful to have the stream type in the generated code (similar to other Connect implementations). In the future we may use this information to optimize stream handling - for example, avoiding the pipe/duplex/oneshot for server streaming calls to enable Connect over HTTP/1.1. Update the generator to include the stream type in the generated code.

Additionally, move the HTTP method to the HTTPRequest class - mutating a method spec to determine which HTTP method to use seems wrong. We should keep MethodSpec to contain only fields set from the generator.


protoc-gen-connect-kotlin/src/main/kotlin/com/connectrpc/protocgen/connect/Generator.kt
```diff
@@ -108,24 +110,25 @@ class Generator : CodeGenerator {
                 .addFileComment("Source: ${file.name}\n")
                 .addType(serviceClientInterface(packageName, service, sourceInfo))
                 .build()
-            fileSpecs.put(serviceClientInterfaceClassName(packageName, service), interfaceFileSpec)
+            fileSpecs[serviceClientInterfaceClassName(packageName, service)] = interfaceFileSpec
 
             val implementationFileSpecBuilder = FileSpec.builder(packageName, file.name)
                 // Manually import `method()` since it is a method and not a class.
```
jhump, line 116: This comment looks stale. Remove?

library/src/test/kotlin/com/connectrpc/impl/BiDirectionalStreamTest.kt
jhump, line 1: I saw this today, but didn't fix it since I didn't otherwise touch this file. But... while you are in here anyway, mind fixing the name? `BiDirectionalStreamTest` -> `BidirectionalStreamTest`? You may have to do something funny to properly do so and have it updated in the git index due to MacOS's case-insensitive file system.

library/src/main/kotlin/com/connectrpc/MethodSpec.kt
```diff
@@ -29,12 +24,12 @@ internal object Method {
...
- * @param method The HTTP method of a request.
+ * @param streamType The method's stream type.
  */
 class MethodSpec<Input : Any, Output : Any>(
     val path: String,
     val requestClass: KClass<Input>,
     val responseClass: KClass<Output>,
     val idempotency: Idempotency = Idempotency.UNKNOWN,
-    val method: String = Method.POST_METHOD,
+    val streamType: StreamType = StreamType.UNKNOWN,
```
jhump, line 34: This unknown value seems a little scary. How will it impact things? Is it necessary for backwards-compatibility with older generated code? What could go wrong if older code is paired with this newer runtime -- would it mean that the GET HTTP verb is never used and interceptors would see the wrong value for stream type? I think it might be safer to make this a non-optional parameter with no default and remove the `UNKNOWN` sentinel enum entry. Users must re-generate their code when they update to newer version of the library.

jhump (approved):
Only one real remark (other two are little nits). Otherwise LG

library/src/main/kotlin/com/connectrpc/MethodSpec.kt
pkwarren: I went back and forth on this. Having UNKNOWN also matches connect-go and grpc-java. connect-swift doesn't have an unknown type.

It was less about the compatibility with existing generated code as this is still pre-v1 and it should be expected that consumers keep the runtime libraries in sync with the generated code.

jhump: I know connect-go has this because there's no real support for enums in Go, and it's just good practice for the default/zero value to be a sentinel of this sort. But in kotlin and java, it's much less clear why you'd include one. With grpc-java, it's likely because that supports server-side, too, and if you want to create an "unknown method handler", it's most correct for it to report to handler/interceptor code that the stream type is unknown.

jhump (approved):
LGTM!

PR #173, opened 2023-12-08
Fix URL resolution for base URI with absolute path

### LINKED ISSUE #149: URL assembly issue
I've noticed a potential issue with the `urlFromMethodSpec` method in the `com.connectrpc.impl.ProtocolClient` class. It seems to have a problem when concatenating the domain and path.

image

For example:
If `baseUri` is `"https://hello.com/api"` and `methodSpec.path` is `"world"`, 

the expected result should be:
`https://hello.com/api/world`, 
but the actual result is:
`https://hello.com/world`. 

The `"api"` part is omitted. I'm wondering if this is a feature or a bug?

pkwarren:
If a client uses a base URI with an absolute path (i.e. 'https://connectrpc.com/api/'), then any RPC methods should be resolved relative to the absolute path (instead of ignoring it as they do today).

Fixes #149.

### DIFF
--- a/library/src/main/kotlin/com/connectrpc/impl/ProtocolClient.kt
+++ b/library/src/main/kotlin/com/connectrpc/impl/ProtocolClient.kt
@@ -34,7 +34,6 @@ import com.connectrpc.protocols.GETConfiguration
 import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.suspendCancellableCoroutine
-import java.net.URL
 import java.util.concurrent.CountDownLatch
 import kotlin.coroutines.resume
 
@@ -268,8 +267,5 @@ class ProtocolClient(
         )
     }
 
-    private fun <Input : Any, Output : Any> urlFromMethodSpec(methodSpec: MethodSpec<Input, Output>): URL {
-        val host = config.baseUri.resolve("/${methodSpec.path}")
-        return host.toURL()
-    }
+    private fun urlFromMethodSpec(methodSpec: MethodSpec<*, *>) = config.baseUri.resolve(methodSpec.path).toURL()
 }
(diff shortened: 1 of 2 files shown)


library/src/main/kotlin/com/connectrpc/impl/ProtocolClient.kt
```diff
@@ -268,8 +267,5 @@ class ProtocolClient(
         )
     }
 
-    private fun <Input : Any, Output : Any> urlFromMethodSpec(methodSpec: MethodSpec<Input, Output>): URL {
-        val host = config.baseUri.resolve("/${methodSpec.path}")
-        return host.toURL()
-    }
+    private fun urlFromMethodSpec(methodSpec: MethodSpec<*, *>) = config.baseUri.resolve(methodSpec.path).toURL()
```
pkwarren, line 270: This does the right thing in my testing:

* baseUri=`https://connectrpc.com`, path=`connectrpc.eliza.v1.ElizaService/Say`, url=`https://connectrpc.com/connectrpc.eliza.v1.ElizaService/Say`
* baseUri=`https://connectrpc.com/`, path=`connectrpc.eliza.v1.ElizaService/Say`, url=`https://connectrpc.com/connectrpc.eliza.v1.ElizaService/Say`
* baseUri=`https://connectrpc.com/api`, path=`connectrpc.eliza.v1.ElizaService/Say`, url=`https://connectrpc.com/connectrpc.eliza.v1.ElizaService/Say`
* baseUri=`https://connectrpc.com/api/`, path=`connectrpc.eliza.v1.ElizaService/Say`, url=`https://connectrpc.com/api/connectrpc.eliza.v1.ElizaService/Say`

pkwarren: Updated to preserve relative path when building URLs, so the 3rd and 4th example above behave the same.

jhump (approved):
LGTM!

PR #182, opened 2023-12-20
Adds new conformance/client module

jhump:
This PR is purely additive. It does not make any changes to the existing container/crosstest-based conformance test targets. It's just starting to add code to eventually replace all of that.

So the old conformance tests are in these modules:
- `conformance/common`
- `conformance/google-java`
- `conformance/google-javalite`

The new ones are in these:
- `conformance/client`
- `conformance/client/google-java`
- `conformance/client/google-javalite`

This doesn't include any actual logic yet. This was a rather substantial PR already, so I figured it would be good to get this approach reviewed before fleshing out the conformance client handler logic (which are all stubbed out in `Client.kt` with TODOs).

To avoid the redundant implementations of the current conformance tests, I'm doing the following:
1. Generate all conformance protos, using _lite_ runtime, into the `conformance/client` module. So the shared logic deals with the lite generated message types, since that's the lowest common denominator. (The full runtime should be able to correctly support lite generate code.)
2. This means that the `conformance/client/google-javalite` module doesn't need any add'l generated code.
3. The `conformance/client/google-java` module, on the other hand, does. So the shared code will convert from generated lite messages to normal/non-lite messages. This uses a `buf.gen.lite.yaml` which has a different package prefix, so both sets of generated messages can co-exist. The dependencies of the non-lite module _exclude_ the javalite protobuf runtime, swapping in the full runtime instead.

This also provides a glimpse of what I'm thinking about for an overhaul of the stream interfaces. In particular, see `BidiStreamClient.BidiStream`, `ServerStreamClient.ServerStream`, and `ClientStreamClient.ClientStream`. Currently, these contain adapters from the existing actual stream interfaces to these newer ones. But I'd like to overhaul the main stream interfaces to look more-or-less like these in the future, so let me know what you think.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,8 @@ generate: $(PROTOC) buildplugin generateconformance generateexamples ## Generate
 .PHONY: generateconformance
 generateconformance: $(PROTOC) buildplugin ## Generate protofiles for conformance tests.
 	buf generate --template conformance/buf.gen.yaml -o conformance conformance/proto
+	buf generate --template conformance/client/buf.gen.yaml -o conformance/client buf.build/connectrpc/conformance:v1.0.0-rc1
+	buf generate --template conformance/client/buf.gen.lite.yaml -o conformance/client buf.build/connectrpc/conformance:v1.0.0-rc1
 
 .PHONY: generateexamples
 generateexamples: $(PROTOC) buildplugin ## Generate proto files for example apps.
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ BIN := .tmp/bin
 CACHE := .tmp/cache
 LICENSE_HEADER_YEAR_RANGE := 2022-2023
 LICENSE_HEADER_VERSION := v1.28.1
+CONFORMANCE_VERSION := v1.0.0-rc1
 PROTOC_VERSION ?= 25.1
 GRADLE_ARGS ?=
 
@@ -74,8 +75,8 @@ generate: $(PROTOC) buildplugin generateconformance generateexamples ## Generate
 .PHONY: generateconformance
 generateconformance: $(PROTOC) buildplugin ## Generate protofiles for conformance tests.
 	buf generate --template conformance/buf.gen.yaml -o conformance conformance/proto
-	buf generate --template conformance/client/buf.gen.yaml -o conformance/client buf.build/connectrpc/conformance:v1.0.0-rc1
-	buf generate --template conformance/client/buf.gen.lite.yaml -o conformance/client buf.build/connectrpc/conformance:v1.0.0-rc1
+	buf generate --template conformance/client/buf.gen.yaml -o conformance/client buf.build/connectrpc/conformance:$(CONFORMANCE_VERSION)
+	buf generate --template conformance/client/buf.gen.lite.yaml -o conformance/client buf.build/connectrpc/conformance:$(CONFORMANCE_VERSION)
 
 .PHONY: generateexamples
 generateexamples: $(PROTOC) buildplugin ## Generate proto files for example apps.
--- a/conformance/client/build.gradle.kts
+++ b/conformance/client/build.gradle.kts
@@ -32,7 +32,7 @@ tasks {
 dependencies {
     implementation(project(":okhttp"))
     implementation(libs.kotlin.coroutines.core)
-    implementation(libs.protobuf.kotlin)
+    implementation(libs.protobuf.kotlinlite)
     implementation(libs.protobuf.javalite)
     implementation(libs.okio.core)
     implementation(libs.okhttp.tls)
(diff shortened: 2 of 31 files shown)


Makefile
```diff
@@ -74,6 +74,8 @@ generate: $(PROTOC) buildplugin generateconformance generateexamples ## Generate
 .PHONY: generateconformance
 generateconformance: $(PROTOC) buildplugin ## Generate protofiles for conformance tests.
 	buf generate --template conformance/buf.gen.yaml -o conformance conformance/proto
+	buf generate --template conformance/client/buf.gen.yaml -o conformance/client buf.build/connectrpc/conformance:v1.0.0-rc1
+	buf generate --template conformance/client/buf.gen.lite.yaml -o conformance/client buf.build/connectrpc/conformance:v1.0.0-rc1
```
pkwarren, line 78: nit: Maybe add a `CONFORMANCE_VERSION` above?

conformance/client/build.gradle.kts
```diff
@@ -0,0 +1,39 @@
...
+            // Generated Kotlin code for protobufs uses RequiresOptIn annotation
+            freeCompilerArgs += "-opt-in=kotlin.RequiresOptIn"
+        }
+    }
+}
+
+dependencies {
+    implementation(project(":okhttp"))
+    implementation(libs.kotlin.coroutines.core)
+    implementation(libs.protobuf.kotlin)
```
pkwarren, line 35: Should this be the kotlin lite dependency?

conformance/client/google-java/build.gradle.kts
```diff
@@ -0,0 +1,55 @@
+buildscript {
+    repositories {
+        mavenCentral()
+        gradlePluginPortal()
+    }
+    dependencies {
+        classpath("gradle.plugin.com.github.johnrengelman:shadow:7.1.2")
```
pkwarren, line 7: If possible we should try to keep all the versions defined in the gradle toml version catalogs.

conformance/client/google-java/build.gradle.kts
```diff
@@ -0,0 +1,55 @@
...
+    main {
+        java {
+            srcDir("build/generated/sources/bufgen")
+        }
+    }
+}
+
+dependencies {
+    implementation(project(":conformance:client")) {
+        exclude(group = "com.google.protobuf", module = "protobuf-javalite")
```
pkwarren, line 46: Should exclude the kotlin lite one too.

conformance/client/google-java/src/main/kotlin/com/connectrpc/conformance/client/java/Main.kt
```diff
@@ -0,0 +1,33 @@
...
+        val invokeStyle = ConformanceClientLoop.parseArgs(args)
+        val client = Client(
+            invokerFactory = ::JavaInvoker,
+            serializationFactory = JavaInvoker::serializationStrategy,
+            invokeStyle = invokeStyle,
+        )
+        ConformanceClientLoop.run(System.`in`, System.out, client)
+    } catch (e: Exception) {
+        e.printStackTrace(System.err)
+        Runtime.getRuntime().exit(1)
```
pkwarren, line 31: If we want we could just let exceptioon bubble up which should print the exception and exit with a non-zero exit code on its own.

conformance/client/src/main/kotlin/com/connectrpc/conformance/client/Client.kt
```diff
@@ -0,0 +1,242 @@
...
+            Protocol.PROTOCOL_GRPC -> NetworkProtocol.GRPC
+            Protocol.PROTOCOL_GRPC_WEB -> NetworkProtocol.GRPC_WEB
+            else -> throw RuntimeException("unsupported protocol: $protocol")
+        }
+    }
+
+    private fun asOkHttpProtocol(httpVersion: HTTPVersion): okhttp3.Protocol {
+        return when (httpVersion) {
+            HTTPVersion.HTTP_VERSION_1 -> okhttp3.Protocol.HTTP_1_1
+            HTTPVersion.HTTP_VERSION_2 -> okhttp3.Protocol.HTTP_2
```
pkwarren, line 170: If we aren't using TLS, we'll need to configure https://square.github.io/okhttp/5.x/okhttp/okhttp3/-protocol/-h2_-p-r-i-o-r_-k-n-o-w-l-e-d-g-e/index.html.

pkwarren (commented):
Looks really nice - the new interfaces look good to me and it will be nice to not have duplicated code.

pkwarren (approved):

pkwarren (approved):
Looks good - just a conflict to resolve in versions toml file after some dependabot updates from yesterday.

pkwarren (approved):

PR #193, opened 2024-01-09
Fixes to get unary conformance test cases passing

jhump:
I had a branch where I got all of the unary test cases running and working with this repo. But it was very large, so I'm breaking it up.

First up are the changes to the actual runtime to get everything to work properly. There were numerous issues with things that were not really covered by the existing crosstest cases. In particular, error details and Connect GET flat-out didn't work. Also, there was no way to plumb extension or type registries through to the codec, so use of `google.protobuf.Any` with JSON format really couldn't work. There were some other little things I noticed and changed along the way. I'll drop a self-review and leave comments to explain each change and what it was trying to accomplish or why it was needed to fix things.

### DIFF
--- a/extensions/google-java/src/main/kotlin/com/connectrpc/extensions/GoogleJavaJSONAdapter.kt
+++ b/extensions/google-java/src/main/kotlin/com/connectrpc/extensions/GoogleJavaJSONAdapter.kt
@@ -64,6 +64,8 @@ internal class GoogleJavaJSONAdapter<E : Message>(
         if (deterministic) {
             printer = printer.sortingMapKeys()
         }
+        // TODO: It would likely be more efficient to use printer.appendTo
+        //       with an Appendable implementation that wraps a Buffer.
         val jsonString = printer.print(message)
         return Buffer().write(jsonString.encodeUtf8())
     }
--- a/extensions/google-java/src/main/kotlin/com/connectrpc/extensions/GoogleJavaJSONAdapter.kt
+++ b/extensions/google-java/src/main/kotlin/com/connectrpc/extensions/GoogleJavaJSONAdapter.kt
@@ -16,54 +16,55 @@ package com.connectrpc.extensions
 
 import com.connectrpc.CODEC_NAME_JSON
 import com.connectrpc.Codec
-import com.google.protobuf.GeneratedMessageV3
 import com.google.protobuf.Internal
+import com.google.protobuf.Message
+import com.google.protobuf.TypeRegistry
 import com.google.protobuf.util.JsonFormat
 import okio.Buffer
 import okio.BufferedSource
 import okio.ByteString.Companion.encodeUtf8
 import kotlin.reflect.KClass
+import kotlin.reflect.cast
 
 /**
  * Adapter for Connect to use Google's protobuf-java runtime for
  * deserializing and serializing data types.
  */
-internal class GoogleJavaJSONAdapter<E : GeneratedMessageV3>(
-    clazz: KClass<E>,
+internal class GoogleJavaJSONAdapter<E : Message>(
+    private val clazz: KClass<E>,
+    private val registry: TypeRegistry,
 ) : Codec<E> {
-    /**
-     * Casting assumes the user is using Google's GeneratedMessageV3 type.
-     */
-    @Suppress("UNCHECKED_CAST")
     private val instance by lazy {
-        Internal.getDefaultInstance(clazz.java as Class<GeneratedMessageV3>)
+        Internal.getDefaultInstance(clazz.java)
     }
 
     override fun encodingName(): String {
         return CODEC_NAME_JSON
     }
 
-    /**
-     * Casting assumes the user is using Google's GeneratedMessageV3 type.
-     * The builder returns a GeneratedMessageV3 but it is assumed to be used
-     * with the assumption that the generic E is the underlying type.
-     */
-    @Suppress("UNCHECKED_CAST")
     override fun deserialize(source: BufferedSource): E {
-        val builder = instance.toBuilder()
-        JsonFormat.parser().ignoringUnknownFields().merge(source.readUtf8(), builder)
-        return builder.build() as E
+        val builder = instance.newBuilderForType()
+        JsonFormat.parser()
+            .ignoringUnknownFields()
+            .usingTypeRegistry(registry)
+            .merge(source.readUtf8(), builder)
+        return clazz.cast(builder.build())
     }
 
     override fun serialize(message: E): Buffer {
-        val jsonString = JsonFormat.printer().print(message)
-        return Buffer().write(jsonString.encodeUtf8())
+        return serialize(message, false)
     }
 
     override fun deterministicSerialize(message: E): Buffer {
-        val jsonString = JsonFormat.printer()
-            .sortingMapKeys()
-            .print(message)
+        return serialize(message, true)
+    }
+
+    private fun serialize(message: E, deterministic: Boolean): Buffer {
+        var printer = JsonFormat.printer()
+        if (deterministic) {
+            printer = printer.sortingMapKeys()
+        }
+        val jsonString = printer.print(message)
         return Buffer().write(jsonString.encodeUtf8())
     }
 }
--- a/extensions/google-java/src/main/kotlin/com/connectrpc/extensions/GoogleJavaJSONStrategy.kt
+++ b/extensions/google-java/src/main/kotlin/com/connectrpc/extensions/GoogleJavaJSONStrategy.kt
@@ -18,25 +18,31 @@ import com.connectrpc.CODEC_NAME_JSON
 import com.connectrpc.Codec
 import com.connectrpc.ErrorDetailParser
 import com.connectrpc.SerializationStrategy
-import com.google.protobuf.GeneratedMessageV3
+import com.google.protobuf.Message
+import com.google.protobuf.TypeRegistry
 import kotlin.reflect.KClass
+import kotlin.reflect.full.isSubclassOf
 
 /**
  * The Google Java JSON serialization strategy.
  */
-class GoogleJavaJSONStrategy : SerializationStrategy {
+class GoogleJavaJSONStrategy(
+    private val registry: TypeRegistry = TypeRegistry.getEmptyTypeRegistry(),
+) : SerializationStrategy {
     override fun serializationName(): String {
         return CODEC_NAME_JSON
     }
 
-    /**
-     * This unchecked cast assumes the underlying class type is
-     * a Google GeneratedMessageV3.
-     */
-    @Suppress("UNCHECKED_CAST")
     override fun <E : Any> codec(clazz: KClass<E>): Codec<E> {
-        val messageClass = clazz as KClass<GeneratedMessageV3>
-        return GoogleJavaJSONAdapter(messageClass) as Codec<E>
+        if (!clazz.isSubclassOf(Message::class)) {
+            throw RuntimeException("class ${clazz.qualifiedName} does not extend MessageLite")
+        }
+        @Suppress("UNCHECKED_CAST") // we just checked above, so it's safe
+        val messageClass = clazz as KClass<out Message>
+
+        @Suppress("UNCHECKED_CAST") // messageClass is actually KClass<E>, so it's safe
+        val adapter = GoogleJavaJSONAdapter(messageClass, registry) as Codec<E>
+        return adapter
     }
 
     override fun errorDetailParser(): ErrorDetailParser {
--- a/extensions/google-java/src/main/kotlin/com/connectrpc/extensions/GoogleJavaJSONStrategy.kt
+++ b/extensions/google-java/src/main/kotlin/com/connectrpc/extensions/GoogleJavaJSONStrategy.kt
@@ -35,7 +35,7 @@ class GoogleJavaJSONStrategy(
 
     override fun <E : Any> codec(clazz: KClass<E>): Codec<E> {
         if (!clazz.isSubclassOf(Message::class)) {
-            throw RuntimeException("class ${clazz.qualifiedName} does not extend MessageLite")
+            throw RuntimeException("class ${clazz.qualifiedName} does not extend Message")
         }
         @Suppress("UNCHECKED_CAST") // we just checked above, so it's safe
         val messageClass = clazz as KClass<out Message>
--- a/extensions/google-java/src/main/kotlin/com/connectrpc/extensions/GoogleJavaProtoAdapter.kt
+++ b/extensions/google-java/src/main/kotlin/com/connectrpc/extensions/GoogleJavaProtoAdapter.kt
@@ -54,6 +54,10 @@ internal class GoogleJavaProtoAdapter<E : MessageLite>(
     }
 
     private fun serialize(message: E, deterministic: Boolean): Buffer {
+        // TODO: It would likely be more efficient to wrap a Buffer with
+        //       an OutputStream implementation so we don't have to
+        //       first create a separate byte array that gets copied to
+        //       buffer.
         val result = ByteArray(message.serializedSize)
         val output = CodedOutputStream.newInstance(result)
         if (deterministic) {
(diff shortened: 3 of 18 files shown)


extensions/google-java/src/main/kotlin/com/connectrpc/extensions/GoogleJavaJSONAdapter.kt
```diff
@@ -16,54 +16,54 @@ package com.connectrpc.extensions
...
 import okio.BufferedSource
 import okio.ByteString.Companion.encodeUtf8
 import kotlin.reflect.KClass
+import kotlin.reflect.cast
 
 /**
  * Adapter for Connect to use Google's protobuf-java runtime for
  * deserializing and serializing data types.
  */
-internal class GoogleJavaJSONAdapter<E : GeneratedMessageV3>(
```
jhump, line 31: Two things: (1) we need to be able to weave a `TypeRegistry` through in order to use `google.protobuf.Any` with the JSON format. And (2) this type is way, too specific. In particular, this wouldn't work with dynamic messages, and there's no particular reason to require this type. So I instead opted for the more generic `Message`, since that is really all that is required.

extensions/google-java/src/main/kotlin/com/connectrpc/extensions/GoogleJavaJSONAdapter.kt
```diff
@@ -16,54 +16,54 @@ package com.connectrpc.extensions
...
-    clazz: KClass<E>,
+internal class GoogleJavaJSONAdapter<E : Message>(
+    private val clazz: KClass<E>,
+    private val registry: TypeRegistry,
 ) : Codec<E> {
-    /**
-     * Casting assumes the user is using Google's GeneratedMessageV3 type.
-     */
-    @Suppress("UNCHECKED_CAST")
```
jhump, line 37: I got rid of unnecessary suppressions. In some cases (like this one), no unchecked cast was even needed/useful. In other places, we can use `clazz.cast` to perform an actual checked cast.

extensions/google-java/src/main/kotlin/com/connectrpc/extensions/GoogleJavaJSONStrategy.kt
```diff
@@ -33,10 +37,16 @@ class GoogleJavaJSONStrategy : SerializationStrategy {
      * This unchecked cast assumes the underlying class type is
      * a Google GeneratedMessageV3.
      */
-    @Suppress("UNCHECKED_CAST")
```
jhump, line 36: I moved the suppressions down to each relevant statement and added a comment as to why it's actually safe. It wasn't previously safe, but now it is because of an added check that the given type is actually a sub-type of the desired type (which I've changed from `GeneratedMessageV3` to `Message` since that is all that is needed and `GeneratedMessageV3` is overly precise and prevents things like use of dynamic messages).

extensions/google-java/src/main/kotlin/com/connectrpc/extensions/GoogleJavaProtoAdapter.kt
```diff
@@ -17,56 +17,49 @@ package com.connectrpc.extensions
...
 import kotlin.reflect.KClass
+import kotlin.reflect.cast
 
 /**
  * Adapter to use Google's protobuf-java runtime for
  * deserializing and serializing data types.
  */
-internal class GoogleJavaProtoAdapter<E : GeneratedMessageV3>(
-    clazz: KClass<E>,
+internal class GoogleJavaProtoAdapter<E : MessageLite>(
```
jhump, line 32: Even though this is the java runtime extension, not javalite, I left the signatures here as `MessageLite` and `ExtensionRegistryLite` to maximize the applicability. `Message` and `ExtensionRegistry` in the non-lite runtime _extend_ `MessageLite` and `ExtensionRegistryLite` respectively, so it's safe. And we don't actually need the types to provide any API beyond what `MessageLite` and `ExtensionRegistryLite` provide.

extensions/google-java/src/main/kotlin/com/connectrpc/extensions/JavaErrorParser.kt
```diff
@@ -31,30 +33,22 @@ internal object JavaErrorParser : ErrorDetailParser {
      * expects a generic type E? and the underlying type is a Google
      * Java Message.
      */
-    @Suppress("UNCHECKED_CAST")
     override fun <E : Any> unpack(any: AnyError, clazz: KClass<E>): E? {
-        val value = any.value.utf8().decodeBase64() ?: any.value
```
jhump, line 36: This was really confusing how some things were stored as `ByteString`, but actually were intended to be base64-encoded strings. This was also a source of problems -- somewhere along the way, things were not correctly decoded, or perhaps were inadvertently getting double-decoded, because error details didn't actually work.

So I've changed it so that the raw JSON model has a base64-encoded string. And every other representation is the decoded bytes. So there's no base64-decoding in this method anymore.

Also, this was building an `Any`, and then deserializing, but it seemed simpler to just directly deserialize from the bytes.

extensions/google-java/src/main/kotlin/com/connectrpc/extensions/GoogleJavaJSONAdapter.kt
jhump: Aside: it's rather annoying that the JSON parser doesn't accept an `ExtensionRegistry`. I double-checked the implementation code and, sure enough, the JSON format in Java does not support parsing extension fields 😢.

extensions/google-java/src/main/kotlin/com/connectrpc/extensions/GoogleJavaJSONAdapter.kt
```diff
@@ -16,54 +16,54 @@ package com.connectrpc.extensions
...
-    private val instance by lazy {
-        Internal.getDefaultInstance(clazz.java as Class<GeneratedMessageV3>)
-    }
+    private val instance: E
+        get() = Internal.getDefaultInstance(clazz.java)
```
jhump, line 38: Hmm... I can no longer recall why I changed this from being a lazy property. I think I need to change it back so it doesn't have to pay the reflection cost every time the property is accessed...

extensions/google-java/src/main/kotlin/com/connectrpc/extensions/JavaErrorParser.kt
```diff
@@ -31,30 +33,22 @@ internal object JavaErrorParser : ErrorDetailParser {
...
-        val unpacked = anyProto.unpack(kClass)
-        if (unpacked?.javaClass?.isAssignableFrom(clazz.java) == true) {
-            return unpacked as E?
+        if (!clazz.isSubclassOf(MessageLite::class)) {
+            throw RuntimeException("class ${clazz.qualifiedName} does not extend MessageLite")
         }
-        return null
+        @Suppress("UNCHECKED_CAST") // we just checked above, so it's safe
+        val instance = Internal.getDefaultInstance(clazz.java as Class<MessageLite>) as MessageLite
+        val unpacked = instance.parserForType.parseFrom(any.value.toByteArray())
```
jhump, line 42: I just realized that one thing we lose with the change to directly deserialize, instead of using `Any.unpack`, is that we're no longer validating the message type indicated in the URL matches the given message class. I'll push a change for that...

extensions/google-java/src/main/kotlin/com/connectrpc/extensions/GoogleJavaProtoAdapter.kt
```diff
@@ -17,56 +17,49 @@ package com.connectrpc.extensions
...
     }
 
     override fun deterministicSerialize(message: E): Buffer {
-        return try {
-            val result = Buffer()
-            val output = CodedOutputStream.newInstance(result.outputStream())
+        return serialize(message, true)
+    }
+
+    private fun serialize(message: E, deterministic: Boolean): Buffer {
```
jhump, line 55: I made both deterministic and non-deterministic flows go through the same flow to make sure they both worked. Previously, the deterministic flow had several serious problems that meant that Connect GET didn't actually work at all:
1) It used `Buffer.outputStream()` to write the marshalled data. However, that method returns the buffer's _underlying buffered sink_, or a no-op stream if there isn't one. We create a new buffer with default constructor, so it has no sink. So the resulting buffers were _always_ empty 😭 
2) It called `output.checkNoSpaceLeft`, but the coded stream was created using an output stream, not a byte array, so this method _always threw an exception_ ([link](https://github.com/protocolbuffers/protobuf/blob/main/java/core/src/main/java/com/google/protobuf/CodedOutputStream.java#L2216-L2218)).

I've left the call to `checkNoSpaceLeft` since now we're creating the output stream with a byte array (which was necessary since `Buffer.outputStream()` doesn't actually do the right thing).

One possible way to improve this and remove this allocation/copy of bytes would be to write an adapter around an `okio.Buffer` that implements [`ByteOutput`](https://github.com/protocolbuffers/protobuf/blob/main/java/core/src/main/java/com/google/protobuf/ByteOutput.java).

extensions/google-javalite/src/main/kotlin/com/connectrpc/extensions/GoogleLiteProtoAdapter.kt
```diff
@@ -17,51 +17,52 @@ package com.connectrpc.extensions
...
     }
 
     override fun deterministicSerialize(message: E): Buffer {
-        return try {
-            val result = Buffer()
-            val output = CodedOutputStream.newInstance(result.outputStream())
+        return serialize(message, true)
+    }
+
+    private fun serialize(message: E, deterministic: Boolean): Buffer {
```
jhump, line 58: Same remark as non-lite java codec -- serious bugs in the deterministic flow, so I consolidated them so the deterministic special-case is more trivially correct.

library/src/main/kotlin/com/connectrpc/ProtocolClientInterface.kt
```diff
@@ -110,3 +110,11 @@ interface ProtocolClientInterface {
         methodSpec: MethodSpec<Input, Output>,
     ): ClientOnlyStreamInterface<Input, Output>
 }
+
+fun Headers.toLowercase(): Headers {
```
jhump, line 114: Some places were using `mapKeys` to turn keys to lower-case. But that is subtly incorrect if two different keys map to the same key after lower-casing, like `content-type` and `Content-Type`. So now we correctly merge the values if there are conflicting mixed-case keys.

This formulation is so that we don't actually need to compute intermediate collections, and basically create and process multiple maps to compute the end result. But maybe an imperative loop would be more clear?

library/src/main/kotlin/com/connectrpc/protocols/ConnectInterceptor.kt
```diff
@@ -278,15 +279,10 @@ internal class ConnectInterceptor(
...
-            val payload = detail.value
-            if (payload == null) {
-                errorDetails.add(ConnectErrorDetail(detail.type, ByteString.EMPTY))
-                continue
-            }
             errorDetails.add(
                 ConnectErrorDetail(
                     detail.type,
-                    payload.encodeUtf8(),
+                    detail.value?.decodeBase64() ?: ByteString.EMPTY,
```
jhump, line 285: This is basically the one place we need to base64-decode the error detail value. From here on out, they're decoded bytes.

library/src/main/kotlin/com/connectrpc/protocols/ConnectInterceptor.kt
```diff
@@ -295,14 +291,11 @@ internal class ConnectInterceptor(
 }
 
 private fun Headers.toTrailers(): Trailers {
-    val trailers = mutableMapOf<String, MutableList<String>>()
-    for (pair in this.filter { entry -> entry.key.startsWith("trailer-") }) {
-        val key = pair.key.substringAfter("trailer-")
-        if (trailers.containsKey(key)) {
-            trailers[key]?.add(pair.value.first())
-        } else {
-            trailers[key] = mutableListOf(pair.value.first())
```
jhump, line 304: I don't know why this used `value.first()`, but it basically meant we only ever kept the first trailer value for a particular key. So if a test case used multiple values for the same key, other after the first were lost. I looked at a couple of formulations, like using the more functional/streaming operators for collections. But this simple imperative for loop was the most concise and simplest to follow.

library/src/main/kotlin/com/connectrpc/protocols/GRPCWebInterceptor.kt
```diff
@@ -244,7 +244,14 @@ internal class GRPCWebInterceptor(
             if (i > 0) {
                 val name = line.substring(0, i).trim()
                 val value = line.substring(i + 1).trim()
-                trailers[name.lowercase()] = listOf(value)
```
jhump, line 247: This failed to merge values, both if there were multiple values for a single key and if there were conflicts after lower-casing. Instead of "last one wins" with a singular value, we merge the values.

okhttp/src/main/kotlin/com/connectrpc/okhttp/ConnectOkHttpClient.kt
```diff
@@ -64,11 +66,7 @@ class ConnectOkHttpClient @JvmOverloads constructor(
             newCall.enqueue(
                 object : Callback {
                     override fun onFailure(call: Call, e: IOException) {
-                        val code = if (e.message?.lowercase() == "canceled") {
-                            Code.CANCELED
-                        } else {
-                            Code.UNKNOWN
-                        }
```
jhump, line 71: We handled "cancelled" in a special way here in the unary flow. We handled "timeout" in a special way over in the stream flow. So I consolidated both checks into a new `codeFromIOException` so now both unary and stream calls can benefit from the classification of canceled and deadline exceeded codes.

jhump: TODO: I can't find anything that actually produces an exception with the word "canceled", so I'm not sure this check really works 🤷. I think most things in the Java ecosystem actually spell it with two Ls, like [`CancellationException`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CancellationException.html)...

extensions/google-java/src/main/kotlin/com/connectrpc/extensions/GoogleJavaJSONStrategy.kt
```diff
@@ -18,25 +18,31 @@ import com.connectrpc.CODEC_NAME_JSON
...
 
-    /**
-     * This unchecked cast assumes the underlying class type is
-     * a Google GeneratedMessageV3.
-     */
-    @Suppress("UNCHECKED_CAST")
     override fun <E : Any> codec(clazz: KClass<E>): Codec<E> {
-        val messageClass = clazz as KClass<GeneratedMessageV3>
-        return GoogleJavaJSONAdapter(messageClass) as Codec<E>
+        if (!clazz.isSubclassOf(Message::class)) {
```
pkwarren, line 37: Should this be `MessageLite`?

pkwarren (approved):
Nice cleanups!

extensions/google-java/src/main/kotlin/com/connectrpc/extensions/GoogleJavaJSONStrategy.kt
jhump: Unfortunately, no. The JSON format requires full messages since it must examine the descriptor to discover the `json_name` option.

pkwarren: Ok then we should update the message on line 38 to use Message instead of MessageLite in the exception.

jhump: Ah, right! Thanks for catching that!

Fixed. I also added a few TODOs based on other comments I left here in a self-review, since in the code seems better than in a PR comment.

pkwarren (approved):

PR #194, opened 2024-01-09
Must use alternate model because of incompatibilities between java and javalite runtimes with well-known types

jhump:
Refresher: the approach made use of lite and non-lite generated code in the same process (generated to different packages), so that the core client could use the lite messages, and the non-lite tests would just swap in the non-lite runtime and then do runtime-conversions to the non-lite message types (by serializing and then de-serializing).

I _thought_ I had done enough with the previous formulation to vet that the approach would work. It compiled fine and even ran fine -- it parsed the first request from the test runner, but then blew up (as expected) from the `TODO` call in `Client.kt`.

Sadly, that was not enough. Trying to unmarshal later test cases would blow up with a class verifier error(!!!). This is because of the way classes are lazy-loaded in the JVM. So it only loaded the `ClientCompatResponse.Cancel` message later, when a request actually _used_ the relevant field. And that class, generated for the lite runtime, has an incompatibility between the lite and non-lite runtimes 😢. The reason is because this message refers to `google.protobuf.Empty`, which extends `google.protobuf.GeneratedMessageLite` in the lite runtime but `GeneratedMessageV3` in the non-lite runtime. Some of the generated code in `Cancel` compiles to a virtual call to a method of `Empty` that is defined on the `GeneratedMessageLite` super-class in the lite runtime. This results in an `invokevirtual` instruction in the byte code that indicates a method of `GeneratedMessageLite`, and that is passing `Empty` as the receiver. When loading that class with the non-lite runtime, the verifier sees that this is invalid because `Empty` does not extend `GeneratedMessageLite`. 🤦

So... back to the drawing board. I've now manually created POJOs for the request and relevant sub-messages, to which the lite and non-lite runtimes can adapt their generated messages. So the generated code for the lite runtime moves out of `conformance/client` and into `conformance/client/google-javalite`, since `conformance/client` now just refers to its own POJO models.

Unfortunately, this means more code -- both to define the POJOs and to implement conversion from the generated Protobuf types to those POJOs, which is _practically_ the same code, but just referencing different types (lite vs. non-lite generated types). Still better than _everything_ being forked across the two runtimes, but not as concise as I was hoping it could be.


conformance/client/google-java/build.gradle.kts
```diff
@@ -16,12 +16,12 @@ plugins {
...
     compileKotlin {
         kotlinOptions {
-            // Generated Kotlin code for protobufs uses OptIn annotation
+            // Generated Kotlin code for protobuf uses OptIn annotation
             freeCompilerArgs += "-opt-in=kotlin.RequiresOptIn"
         }
     }
     shadowJar {
-        archiveBaseName.set("shadow")
+        archiveFileName.set("conformance-client-java.jar")
```
jhump, line 24: I made this change to make it easier to actually use the output JAR from another target in the `Makefile`. Without this, the actual file has a `<version>-SNAPSHOT` in the filename. But I couldn't find any simple way to figure out the `<version>`, so that a command could reference the correct filename. (It appears to be computed dynamically by Gradle based on git tags.) So this removes the suffix so we can easily reference the correct output JAR file.

Makefile
```diff
@@ -10,7 +10,7 @@ BIN := .tmp/bin
 CACHE := .tmp/cache
 LICENSE_HEADER_YEAR_RANGE := 2022-2023
 LICENSE_HEADER_VERSION := v1.28.1
-CONFORMANCE_VERSION := v1.0.0-rc1
+CONFORMANCE_VERSION := v1.0.0-rc2
```
jhump, line 13: The main change this brings in is a new method on the conformance service: `IdempotentUnary`, which is just like `Unary`, but it is declared to be side-effect-free so it can be used with Connect GET.

conformance/client/google-java/src/main/kotlin/com/connectrpc/conformance/client/java/JavaInvoker.kt
```diff
@@ -49,14 +50,4 @@ class JavaInvoker(
     override fun bidiStreamClient(): BidiStreamClient<*, *> {
         return JavaBidiStreamClient(client)
     }
-
-    companion object {
-        fun serializationStrategy(codec: Codec): SerializationStrategy {
```
jhump, line 54: I moved this helper function into the new `JavaHelpers`, which contains other helper functions now that there are more things to adapt from the generated code to the model used by `conformance/client`.

conformance/client/google-java/src/main/kotlin/com/connectrpc/conformance/client/java/Main.kt
```diff
@@ -18,16 +18,17 @@ import com.connectrpc.conformance.client.Client
...
-        val invokeStyle = ConformanceClientLoop.parseArgs(args)
-        val client = Client(
-            invokerFactory = ::JavaInvoker,
-            serializationFactory = JavaInvoker::serializationStrategy,
-            invokeStyle = invokeStyle,
-        )
-        ConformanceClientLoop.run(System.`in`, System.out, client)
-    } catch (e: Exception) {
-        e.printStackTrace(System.err)
-        Runtime.getRuntime().exit(1)
```
jhump, line 31: Oops. You left a comment about this in the previous PR, but I mistakenly only applied it to the `google-javalite/.../Main.kt` file and not to this one. So correcting that here.

conformance/client/src/main/kotlin/com/connectrpc/conformance/client/Client.kt
```diff
@@ -49,56 +47,104 @@ import java.security.spec.PKCS8EncodedKeySpec
...
+                CLIENT_STREAM_METHOD_NAME -> handleClient(invoker.clientStreamClient(), req)
+                SERVER_STREAM_METHOD_NAME -> handleServer(invoker.serverStreamClient(), req)
+                BIDI_STREAM_METHOD_NAME -> handleBidi(invoker.bidiStreamClient(), req)
+                else -> throw RuntimeException("method ${req.method} is not known")
             }
-            else -> throw RuntimeException("method $method is not known")
+        } finally {
+            // Clean-up HTTP client.
+            httpClient.connectionPool.evictAll()
+            httpClient.dispatcher.executorService.shutdown()
```
jhump, line 96: I made these changes so that the conformance client can cleanly shutdown without using `System.exit`. It turns out that the old code was creating lots of HTTP clients and never cleaning them up, which left non-daemon threads around, preventing the JVM from promptly terminating after the `main` function returned.

conformance/client/src/main/kotlin/com/connectrpc/conformance/client/Client.kt
```diff
@@ -49,56 +47,104 @@ import java.security.spec.PKCS8EncodedKeySpec
...
         }
     }
 
-    private suspend fun <Req : MessageLite, Resp : MessageLite> handleUnary(
+    private suspend fun <
+        Req : MessageLite,
+        Resp : MessageLite,
+        > handleUnary(
         client: UnaryClient<Req, Resp>,
+        requestType: String,
```
jhump, line 105: This parameter isn't used yet, but it will be needed since the lite runtime generated code doesn't actually know the protobuf names for generated types, and we need to verify that the request indicates the correct request type (since this method is re-used across all unary operations, so three different request message types: `UnaryRequest`, `IdempotentUnaryRequest`, and `UnimplementedRequest`).

conformance/client/src/main/kotlin/com/connectrpc/conformance/client/Client.kt
```diff
@@ -113,63 +159,66 @@ class Client(
...
-            HTTPVersion.HTTP_VERSION_2 -> if (useTls) okhttp3.Protocol.HTTP_2 else okhttp3.Protocol.H2_PRIOR_KNOWLEDGE
-            else -> throw RuntimeException("unsupported HTTP version: $httpVersion")
+            HttpVersion.HTTP_1_1 -> listOf(okhttp3.Protocol.HTTP_1_1)
+            HttpVersion.HTTP_2 ->
+                if (useTls) {
+                    // okhttp *requires* that protocols contains HTTP_1_1
+                    // or H2_PRIOR_KNOWLEDGE. So we leave 1.1 in here, but
+                    // expect HTTP/2 to always be used in practice since it
+                    // should be negotiated during TLS handshake,
+                    listOf(okhttp3.Protocol.HTTP_2, okhttp3.Protocol.HTTP_1_1)
```
jhump, line 218: The previous formulation could return a set of protocols that _only_ included `HTTP_2`. However, `okhttp3` doesn't like this and expects `HTTP_1_1` or `H2_PRIOR_KNOWLEDGE` to also be in the list, just in case it encounters a non-TLS request URL.

jhump (commented):
This PR also adds some more doc comments to the public classes in `conformance/client`.

conformance/client/google-java/build.gradle.kts
pkwarren: Another option here is that you can make it use the application plugin (https://docs.gradle.org/current/userguide/application_plugin.html) which when combined with `installDist` will create an executable script with the right jar file name embedded. I do that in protoc-gen-connect-kotlin to make it callable.

pkwarren (approved):
Overall looks good. I like that we're testing both lite and full runtimes, but it feels pretty complex to go through all the hoops. I guess another alternative would be we could just run all the tests on the lite runtime (since the full runtime is additive) and just write some selected unit tests for the full runtime classes to verify serialization/deserialization.

I'm ok with this approach just wanted to throw out another option if this gets too tedious.

jhump:
> I guess another alternative would be we could just run all the tests on the lite runtime (since the full runtime is additive) and just write some selected unit tests for the full runtime classes to verify serialization/deserialization.

The JSON format is not supported in the lite runtime, so I think it's worth running the whole conformance suite against both runtimes just to make sure we're running the tests using JSON. But I suppose I could be convinced the other way. How about we see how it goes maintaining both, and if maintenance/fixes/etc turn out to be's a pain, we can change our mind and simplify?

conformance/client/google-java/build.gradle.kts
jhump: 🤦, doh! I should have done that to begin with -- good idea! I wasn't aware of the application plugin and was mainly used to building "fat jars" when doing Java dev with Maven way back in the day.

I ripped out the shadowjar stuff and converted both of these conformance programs to use the application plugin. Thanks for the pointer!

PR #195, opened 2024-01-10
Implement unary operations in conformance client

jhump:
This also adds support for other command-line args, to enable verbose tracing output (which was handy for troubleshooting stuff).

Other repos use the make target `runconformance`, but this repo was using `conformancerun`, so I changed it. That target still runs the old cross test stuff, but it also now runs these new unary conformance test cases. The config file is what indicates unary-only. The "known failing" file is because the assertions in the test runner for Connect GET requests are too strict -- AFAICT, this client now passes those tests. The only failing assertion is about the `message` query param that the server gets because it is expecting the encoding in a particular way, but that is not necessary (but requires a change in the test runner to relax the assertion).

The `Makefile` downloads the conformance test runner in the same fashion as it downloads `protoc`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,8 @@ LICENSE_HEADER_VERSION := v1.28.1
 CONFORMANCE_VERSION := v1.0.0-rc2
 PROTOC_VERSION ?= 25.1
 GRADLE_ARGS ?=
+PROTOC := $(BIN)/protoc
+CONNECT_CONFORMANCE := $(BIN)/connectconformance
 
 UNAME_OS := $(shell uname -s)
 UNAME_ARCH := $(shell uname -m)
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -53,18 +88,35 @@ PROTOC_OS = linux
...
+	chmod u+w $@
+
+CONNECT_CONFORMANCE_DOWNLOAD := $(CACHE)/connect-conformance-$(CONFORMANCE_VERSION).tgz
+$(CONNECT_CONFORMANCE_DOWNLOAD):
+	@if ! command -v curl >/dev/null 2>/dev/null; then echo "error: curl must be installed" >&2; exit 1; fi
+	@if ! command -v tar >/dev/null 2>/dev/null; then echo "error: tar must be installed" >&2; exit 1; fi
+	$(eval CONFORMANCE_TMP := $(shell mktemp -d))
+	curl -sSL https://github.com/connectrpc/conformance/releases/download/$(CONFORMANCE_VERSION)/connectconformance-$(CONFORMANCE_VERSION)-$(UNAME_OS)-$(UNAME_ARCH).tar.gz -o $(CONFORMANCE_TMP)/conformance.tgz
 	@mkdir -p $(dir $@)
-	@touch $@
```
jhump, line 67: This seemed weird that it would create an empty placeholder zip file. So I tweaked this so that the artifact it creates is the actual downloaded ZIP file.

I also added a distinct target for the `.tmp/bin/protoc` binary itself because, TBH, I found it very confusing that the `$(PROTOC)` target was actually referencing a bogus ZIP file.

pkwarren: At some point I'd love to move all of this into the gradle build so we only have to maintain a single build system but this is a nice improvement.

conformance/client/src/main/kotlin/com/connectrpc/conformance/client/ClientArgs.kt
```diff
@@ -0,0 +1,76 @@
...
+                    else -> {
+                        if (arg.startsWith("-")) {
+                            throw RuntimeException("unknown option: $arg")
+                        } else {
+                            throw RuntimeException("this command does not accept positional arguments")
+                        }
+                    }
+                }
+            }
+            return ClientArgs(invokeStyle, verbosity)
```
pkwarren, line 73: This works for now but in the future I've had good success with https://args4j.kohsuke.org/ if we need more complexity here.

conformance/client/src/main/kotlin/com/connectrpc/conformance/client/ConformanceClientLoop.kt
```diff
@@ -31,16 +30,27 @@ import java.io.OutputStream
...
     private val requestUnmarshaller: (ByteArray) -> ClientCompatRequest,
     private val responseMarshaller: (ClientCompatResponse) -> ByteArray,
+    private val verbosity: Int = 0,
 ) {
     fun run(input: InputStream, output: OutputStream, client: Client) = runBlocking {
         // TODO: issue RPCs in parallel
         while (true) {
             var result: ClientCompatResponse.Result
             val req = readRequest(input) ?: return@runBlocking // end of stream
+            if (verbosity > 0) {
```
pkwarren, line 40: Another option for more debugging might be to configure the okhttp logging interceptor (perhaps at higher verbosity?).

pkwarren (approved):

conformance/client/src/main/kotlin/com/connectrpc/conformance/client/ClientArgs.kt
jhump: Yeah, I've used a similar library, JCommander. I actually briefly looked into Clickt, which seems to be the most popular library for Kotlin command-line apps. But I had an issue with it (likely something quirky in the IDE, though I did update the project after adding a dep in gradle, but it still wouldn't work), so I decided to punt on pulling in a dep for this, since this logic is so simple 🤷

conformance/client/src/main/kotlin/com/connectrpc/conformance/client/ConformanceClientLoop.kt
jhump: Ack. That would likely be much more verbose than this. I haven't needed that sort of output so far, but it does seem like a potentially useful idea.

PR #196, opened 2024-01-12
Implement stream operations in the conformance client

jhump:
I've got each stream type in its own commit, to hopefully make this easier to review.

We still don't run the conformance tests for stream operations in CI because there are some flaky failures that I need to troubleshoot. In some circumstances, the reference servers are sending back unexpected errors for server stream RPCs, so I'll need to use a debugger with the reference server Go code to figure it out.

I want to dig a _little bit_ more into the timeout issues with bidi RPCs, but I think this is a real bug in okhttp and, if so, will file an issue.


conformance/client/lite-unary-config.yaml
```diff
@@ -20,5 +20,3 @@ features:
     # so that we can run them all three ways: suspend,
     # callback, and blocking.
     - STREAM_TYPE_UNARY
-  supportsTlsClientCerts: true
-  supportsHalfDuplexBidiOverHttp1: true
```
jhump, line 24: I've removed the line about client certs since this property defaults to false. An error is occurring when this code tries to instantiate a client cert from the data in the conformance request. I haven't dug into it, and decided to just omit it for now.

It turns out that we actually _can't_ do half-duplex stuff over HTTP 1.1, because we don't actually know ahead of time, for a given bidi RPC, if it's half-duplex or full-duplex (in the framework, based on generated protobuf metadata). But the request body _must_ be marked as "duplex" if it _might be_ full-duplex. So have to mark the body as duplex for all bidi RPCs, and okhttp then disallows them to be used with HTTP 1.1.

conformance/client/src/main/kotlin/com/connectrpc/conformance/client/adapt/BidiStreamClient.kt
```diff
@@ -44,20 +44,18 @@ abstract class BidiStreamClient<Req : MessageLite, Resp : MessageLite>(
      * @param Resp The response message type
      */
     interface BidiStream<Req : MessageLite, Resp : MessageLite> {
-        fun requests(): RequestStream<Req>
-        fun responses(): ResponseStream<Resp>
+        val requests: RequestStream<Req>
+        val responses: ResponseStream<Resp>
```
jhump, line 48: I've changed a few things in these interfaces to be property getters instead of methods since they really are lightweight properties.

conformance/client/src/main/kotlin/com/connectrpc/conformance/client/adapt/ClientStreamClient.kt
```diff
@@ -46,6 +46,7 @@ abstract class ClientStreamClient<Req : MessageLite, Resp : MessageLite>(
     interface ClientStream<Req : MessageLite, Resp : MessageLite> {
         suspend fun send(req: Req)
         suspend fun closeAndReceive(): ResponseMessage<Resp>
+        suspend fun cancel()
```
jhump, line 49: I had to add this in order to expose the ability to cancel a client-stream RPC. I could have also called it `close` for consistency with other methods. What do you think? Should I change it?

TBH, I kind of _don't like_ that the methods that cancel the RPC are named like "receive close" / "close response", since that's not very clear as to what they're really doing. Maybe when overhauling these interfaces (more-or-less making them look like the interfaces in the conformance client's `adapt` package), we can revisit some of the names and name this operation `cancel` instead of `close` 🤷.

library/src/main/kotlin/com/connectrpc/Code.kt
```diff
@@ -71,7 +74,8 @@ enum class Code(val codeName: String, val value: Int) {
             if (value == null) {
                 return UNKNOWN
             }
-            return values().first { code -> code.value == value }
```
jhump, line 74: This would throw if the given value was not known, which seemed wrong.

okhttp/src/main/kotlin/com/connectrpc/okhttp/OkHttpStream.kt
```diff
@@ -175,7 +172,8 @@ private class ResponseCallback(
     }
 }
 
-internal class PipeDuplexRequestBody(
+internal class PipeRequestBody(
+    private val duplex: Boolean,
```
jhump, line 176: This was necessary in order to do client or server streaming over HTTP 1.1.

conformance/client/lite-unary-config.yaml
pkwarren: Do we want to comment these properties out with a `TODO` to know to come back to them instead of removing them? Up to you but might help identify remaining work since it sounds like there are several things to track down.

conformance/client/src/main/kotlin/com/connectrpc/conformance/client/adapt/ClientStreamClient.kt
pkwarren: I think this is fine for now - maybe there's a better API we could put together that would hook into coroutine cancelation instead of having to expose separate methods for this.

okhttp/src/main/kotlin/com/connectrpc/okhttp/OkHttpStream.kt
pkwarren: Nice! I expected this might be a bit more tricky - especially for client streaming calls.

pkwarren (approved):

conformance/client/src/main/kotlin/com/connectrpc/conformance/client/adapt/ClientStreamClient.kt
jhump: I think this is already hooked into coroutine cancellation:
* [unary](https://github.com/connectrpc/connect-kotlin/blob/main/library/src/main/kotlin/com/connectrpc/impl/ProtocolClient.kt#L140-L147)
* [streaming](https://github.com/connectrpc/connect-kotlin/blob/main/library/src/main/kotlin/com/connectrpc/impl/ProtocolClient.kt#L258-L260)

But, I'm not sure I 100% understand exactly how that actually works. I don't think that alone does what we want it to do: the continuation in question for streaming simply returns the bidirectional stream object. So I think the only thing that this would cancel is the actual call to the streaming RPC method. Once it returns the stream object, that continuation is complete and the cancellation no longer does anything.

So I guess we'd have to do something similar in the app code, to tie its coroutine's cancellation to cancelling the stream (which I think means we still need a `cancel` method of some sort). I think it would be challenging to somehow have the framework automatically manage coroutine lifetimes and tie them to the stream since ownership is unclear -- i.e. it's possible that one coroutine could pass the stream to another that runs on a different thread/worker, in which case it's unclear what coroutine cancellation semantics _should_ be.

PR #218, opened 2024-02-06
Auto-dispatch to I/O context

jhump:
This adds a new property to `ProtocolClientConfig` called `ioCoroutineContext`. It defaults to null, which means there is no change in behavior from the current defaults -- operations happen in the calling coroutine context, so callers of RPC methods must handle the dispatch.

But when it is configured as non-null (most likely using a value like `Dispatchers.IO`), then I/O and blocking operations for an RPC are automatically dispatched to the given coroutine context. This allows for RPC client code to be used from any coroutine context without having to worry about explicitly dispatching.

### DIFF
--- a/examples/android/src/main/kotlin/com/connectrpc/examples/android/ElizaChatActivity.kt
+++ b/examples/android/src/main/kotlin/com/connectrpc/examples/android/ElizaChatActivity.kt
@@ -86,6 +86,9 @@ class ElizaChatActivity : AppCompatActivity() {
                 host = host,
                 serializationStrategy = GoogleJavaLiteProtobufStrategy(),
                 networkProtocol = selectedNetworkProtocolOption,
+                // RPC operations that involve network I/O will
+                // use this coroutine context.
+                ioCoroutineContext = Dispatchers.IO,
             ),
         )
         // Create the Eliza service client.
@@ -113,7 +116,7 @@ class ElizaChatActivity : AppCompatActivity() {
             adapter.add(MessageData(sentence, false))
             editTextView.setText("")
             // Ensure IO context for unary requests.
-            lifecycleScope.launch(Dispatchers.IO) {
+            lifecycleScope.launch {
                 // Make a unary request to Eliza.
                 val response = elizaServiceClient.say(SayRequest.newBuilder().setSentence(sentence).build())
                 response.success { success ->
@@ -133,7 +136,7 @@ class ElizaChatActivity : AppCompatActivity() {
 
     private fun setupStreamingChat(elizaServiceClient: ElizaServiceClient) {
         // On stream result, this callback can be called multiple times.
-        lifecycleScope.launch(Dispatchers.IO) {
+        lifecycleScope.launch {
             // Initialize a bidi stream with Eliza.
             val stream = elizaServiceClient.converse()
             try {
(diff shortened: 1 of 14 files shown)


library/src/main/kotlin/com/connectrpc/http/HTTPClientInterface.kt
```diff
@@ -46,92 +45,3 @@ interface HTTPClientInterface {
      */
     fun stream(request: HTTPRequest, duplex: Boolean, onResult: suspend (StreamResult<Buffer>) -> Unit): Stream
 }
-
-interface Stream {
```
jhump, line 50: Minor cleanup in here: I hoisted the `Stream` stuff out to a separate `Stream.kt` file.

library/src/main/kotlin/com/connectrpc/impl/ProtocolClient.kt
```diff
@@ -172,29 +188,29 @@ class ProtocolClient(
...
-        val stream = stream(headers, methodSpec)
+        val stream = newStream(headers, methodSpec)
         return ClientOnlyStream(stream)
     }
 
-    private suspend fun <Input : Any, Output : Any> bidirectionalStream(
-        methodSpec: MethodSpec<Input, Output>,
+    private fun <Input : Any, Output : Any> newStream(
         headers: Headers,
-    ): BidirectionalStream<Input, Output> = suspendCancellableCoroutine { continuation ->
```
jhump, line 197: More cleanup: the `suspendCancellableCoroutine` usage here didn't really do anything.

It tried to associate cancellation of the stream with the coroutine being cancelled. But this only does anything for the duration of this method body, and only when the coroutine is suspended. 

Since the method always immediately resumes the continuation, this didn't really do anything in practice. So I've removed it for clarity.

Now, it's no longer a suspendable function (so can't be cancelled while executing anyway). Instead, the previous PR (#213, that adds auto-closing block-based RPC invocations to the conformance client) demonstrates the better (IMO) way to go for making sure that the RPCs are appropriately cancelled when a coroutine gets cancelled.

examples/android/src/main/kotlin/com/connectrpc/examples/android/ElizaChatActivity.kt
```diff
@@ -156,15 +159,13 @@ class ElizaChatActivity : AppCompatActivity() {
...
-            lifecycleScope.launch(Dispatchers.Main) {
-                buttonView.setOnClickListener {
-                    val sentence = editTextView.text.toString()
-                    adapter.add(MessageData(sentence, false))
-                    editTextView.setText("")
-                    // Send will be streaming a message to Eliza.
-                    lifecycleScope.launch(Dispatchers.IO) {
-                        stream.send(ConverseRequest.newBuilder().setSentence(sentence).build())
-                    }
+            buttonView.setOnClickListener {
```
pkwarren, line 162: Should code that deals with the UI still be dispatched to Dispatchers.Main? I'm not familiar with android development but we might want to keep this here.

pkwarren (approved):
Changes look good - one question on Android example. I like that this simplifies the other examples.

examples/android/src/main/kotlin/com/connectrpc/examples/android/ElizaChatActivity.kt
jhump: The documentation for [`LifecycleCoroutineScope`](https://developer.android.com/reference/kotlin/androidx/lifecycle/LifecycleCoroutineScope) (the type returned by `lifecycleScope`) states that it is already tied to `Dispatchers.Main`. So the above call to `lifecycleScope.launch` should already dispatch in the right context.

I can set it explicitly above if you'd prefer. I wasn't really sure what was the most idiomatic for Android apps -- to omit when using the default or to always specify explicitly. I guess, even if it is idiomatic to omit, we might want to state it explicitly, just for maximum clarity.

library/src/main/kotlin/com/connectrpc/ProtocolClientConfig.kt
```diff
@@ -45,6 +46,14 @@ class ProtocolClientConfig @JvmOverloads constructor(
     // Compression pools that provide support for the provided `compressionName`, as well as any
     // other compression methods that need to be supported for inbound responses.
     compressionPools: List<CompressionPool> = listOf(GzipCompressionPool),
+    // The coroutine context to use for I/O, such as sending RPC messages.
+    // If null, the current/calling coroutine context is used. So the caller
+    // may need explicitly dispatch send calls using contexts where I/O is
+    // appropriate (using the withContext extension function). If non-null
+    // (such as Dispatchers.IO), operations that involve I/O pr other
+    // blocking will automatically be dispatched using the biven context,
+    // so the caller does not need to worry about it.
```
jhump, line 55:
```suggestion
    // The coroutine context to use for I/O, such as sending RPC messages.
    // If null, the current/calling coroutine context is used. So the caller
    // may need to explicitly dispatch send calls using contexts where I/O
    // is appropriate (using the withContext extension function). If non-null
    // (such as Dispatchers.IO), operations that involve I/O or other
    // blocking will automatically be dispatched using the given context,
    // so the caller does not need to worry about it.
```

pkwarren (approved):

PR #224, opened 2024-02-21
Denormalize "Code", slight simplification of HTTP interface

jhump:
Sorry that this is kind of a large PR. The main intent was the first bullet, but I tackled some other things that I think are improvements along the way.

1. This denormalizes the presence of code. It was previously present both on `StreamResult.Complete` and `ResponseMessage`, even though it was redundant with the code on any non-null `ConnectException`.
2. Removes `Code.OK`. It is now just an error code. An "OK" result is easily distinguished by the lack of an exception.
3. The `HTTPClientInterface` was previously responsible for translating HTTP status -> `Code` and the `HTTPResponse` reported back the status in an optional "trace info" field. But translating status to `Code` should be the responsibility of the protocol implementations, not the HTTP client. (Though it is still up to the HTTP client to classify exceptions that prevent an HTTP response from being received, as these could be client-implementation-specific exception types). So this change promotes the status code out of trace info and into the `HTTPResponse`. 
4. `ProtocolClient.stream` used `suspendCancellableCoroutine` incorrectly. It looks like it was trying to associate the stream with the coroutine, so if the coroutine is cancelled then the stream is closed. But it is only effective *while that method is running*, which is almost none at all since it immediately returns a result (no suspensions).

There are some other small clean-ups in here, too, like moving some things around to places that (IMO) make more sense.

### DIFF
--- a/conformance/client/src/main/kotlin/com/connectrpc/conformance/client/Client.kt
+++ b/conformance/client/src/main/kotlin/com/connectrpc/conformance/client/Client.kt
@@ -386,9 +386,6 @@ class Client(
     private fun unaryResult(numUnsent: Int, result: ResponseMessage<out MessageLite>): ClientResponseResult {
         return when (result) {
             is ResponseMessage.Success -> {
-                if (result.code != Code.OK) {
-                    throw RuntimeException("RPC was successful but ended with non-OK code ${result.code}")
-                }
                 ClientResponseResult(
                     headers = result.headers,
                     payloads = listOf(payloadExtractor(result.message)),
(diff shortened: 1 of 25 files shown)


pkwarren (approved):
Really nice cleanup - I like the removal of the OK code and having to have it on success messages. This brings this implementation into closer parity with connect-go.

erawhctim:
@jhump Just an FYI, this bit threw me off for a sec as we're upgrading to 0.5.0. It could be valuable to call this out explicitly in the release notes 🙏 

> Removes Code.OK. It is now just an error code. An "OK" result is easily distinguished by the lack of an exception.

PR #226, opened 2024-02-23
Update to v1.0.0-rc3 of the conformance suite

jhump:
These new test cases found some uncaught-exception bugs that are fixed in here. It also found some other small bugs in handling anomalous/erroneous responses. These are not high priority, so instead of trying to fix all of them, I've documented them all in the known failing config files.

Here are relevant release notes for rc3: https://github.com/connectrpc/conformance/releases/tag/v1.0.0-rc3

There were some backwards-incompatible changes that required some small changes in the config and the conformance client code. (Once we hit v1.0.0 of the conformance suite, we'll consider it stable and avoid such breaking changes thereafter.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ BIN := .tmp/bin
 CACHE := .tmp/cache
 LICENSE_HEADER_YEAR_RANGE := 2022-2023
 LICENSE_HEADER_VERSION := v1.28.1
-CONFORMANCE_VERSION := v1.0.0-rc2
+CONFORMANCE_VERSION := v1.0.0-rc3
 PROTOC_VERSION ?= 25.3
 GRADLE_ARGS ?=
 PROTOC := $(BIN)/protoc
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -125,13 +125,20 @@ $(CONNECT_CONFORMANCE): $(CONNECT_CONFORMANCE_DOWNLOAD)
 
 .PHONY: generate
 generate: $(PROTOC) buildplugin generateconformance generateexamples ## Generate proto files for the entire project.
+	rm -rf protoc-gen-connect-kotlin/build/generated/sources/bufgen || true
```
jhump, line 128: I added these cleanup calls because we removed an unused message from the conformance protos, and it was causing a compile error until I did this due to an old, invalid source file being left around in the output folder.

conformance/client/src/main/kotlin/com/connectrpc/conformance/client/Client.kt
```diff
@@ -228,17 +228,8 @@ class Client(
...
-                } else {
-                    ConnectException(
-                        code = Code.UNKNOWN,
-                        message = ex.message,
-                        exception = ex,
-                    )
-                }
                 return ClientResponseResult(
-                    error = connEx,
+                    error = asConnectException(ex),
```
jhump, line 232: I consolidated a few `if (ex is ConnectException) ...` blocks with a new helper function.

library/src/main/kotlin/com/connectrpc/impl/ProtocolClient.kt
```diff
@@ -94,7 +96,20 @@ class ProtocolClient(
             val unaryFunc = config.createInterceptorChain()
             val finalRequest = unaryFunc.requestFunction(unaryRequest)
             val cancelable = httpClient.unary(finalRequest) httpClientUnary@{ httpResponse ->
-                val finalResponse = unaryFunc.responseFunction(httpResponse)
+                val finalResponse: HTTPResponse
+                try {
+                    finalResponse = unaryFunc.responseFunction(httpResponse)
+                } catch (ex: Throwable) {
```
jhump, line 102: This was one of the previously uncaught exceptions. It was happening when there was a problem in the wire format of the 5-byte envelopes around stream messages. Instead of being turned into an RPC failure, it was causing an okhttp worker thread to die, which would then cause the conformance client to hang and timeout.

library/src/main/kotlin/com/connectrpc/impl/ProtocolClient.kt
```diff
@@ -129,8 +144,16 @@ class ProtocolClient(
                 )
             }
             return cancelable
-        } catch (e: Exception) {
-            throw e
```
jhump, line 133: This was the other uncaught exception. 🤦
Since this is an async operation, instead of throwing, it needs to invoke the callback with an error.

pkwarren (approved):

PR #232, opened 2024-03-06
Remove old crosstest stuff

jhump:
The new conformance tests are more than adequate replacements. So we can ditch the old tests now.


jhump:
cc @drice-buf, before the current conformance suite, we had an approach that required clients to implement all of the test  cases themselves (instead of a generic client that was data-driven with test case input) and then run those against a large number of server implementations that were distributed via Docker containers. This pull request rips out the old stuff. When I first added the new conformance suite, I left the old one in place, to let them run in parallel. But the new tests have identified a number of issues and are far more thorough, so there's no reason to keep running the old stuff.

pkwarren (approved):

PR #243, opened 2024-03-18
Update to protobuf 4.26.0

pkwarren:
Update to latest protobuf 4.26.0 and regenerate code with the new major version.


jhump (approved):
LGTM.

Aside: I just realized that it's kinda strange that we have to download protoc at all here, since we use `buf` CLI for code gen. I know it's because protoc includes the code generators for the core languages, including Java. But it seems like we could/should use remote plugin execution in this repo's `buf.gen.yaml` files. WDYT?

pkwarren:
> Aside: I just realized that it's kinda strange that we have to download protoc at all here, since we use `buf` CLI for code gen. I know it's because protoc includes the code generators for the core languages, including Java. But it seems like we could/should use remote plugin execution in this repo's `buf.gen.yaml` files. WDYT?

Opened https://github.com/connectrpc/connect-kotlin/issues/244 to migrate to remote plugins.

PR #248, opened 2024-04-01
Update to rc4 of conformance suite and fix several issues

jhump:
There are a handful of changes in here that go along with this update of the conformance tests.

1. This updates the HTTP code -> RPC code mappings per latest updates to the spec [here](https://github.com/connectrpc/connectrpc.com/pull/130).

2. The newest release candidate of conformance tests really wants clients to report "exception metadata" as trailers, and for that to be a combined view of headers _and_ trailers for unary and client-stream errors.

   The reasoning is related to the fact that, in the gRPC and gRPC-Web protocols, a server can choose to combine headers and trailers into a "trailers-only" response when there are no response messages (which is always the case for unary and client-stream error responses). So the caller only checking headers or trailers when querying a particular key is brittle; and the caller checking _both_ headers and trailers is annoying and verbose. Instead, the combined metadata is reported as part of the exception, and the caller then has one bucket of metadata to query for a particular key, which is less error-prone.

3. Unary and client-stream operations now properly handle cases where there is either zero or more than one response message combined with an non-error status. They return "unimplemented" as indicated by [this doc](https://grpc.github.io/grpc/core/md_doc_statuscodes.html) on RPC status codes (search therein for "cardinality violation").

4. In the Connect protocol, we correctly default to the code inferred from the HTTP status code, instead of always falling back to "unknown" in the event that the code cannot be successfully parsed from the JSON error body.

5. The `ConnectException` now uses a private constructor to provide a strong invariant that the error detail parser is always non-nil if there are non-empty error details. Previously, the old constructor allowed a caller to provide non-empty details but a null error parser, in which case the `unpackedDetails` method would not behave correctly. It also had an awkward public `setErrorParser` method which was actually only called from one place and was not even needed (and was technically not needed since it was previously declared as a `data` class, which means it gets an automatic `copy` method that could have been used to set the parser field).

   This is technically a backwards-incompatible change since the public constructor now has two fewer parameters. It is also no longer a `data` class (since the automatic `copy` method would have allowed the new invariant to be subverted -- a caller could invoke `ex.copy(errorDetailParser = null)`).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ BIN := .tmp/bin
 CACHE := .tmp/cache
 LICENSE_HEADER_YEAR_RANGE := 2022-2023
 LICENSE_HEADER_VERSION := v1.30.0
-CONFORMANCE_VERSION := v1.0.0-rc3
+CONFORMANCE_VERSION := v1.0.0-rc4
 PROTOC_VERSION ?= 26.1
 GRADLE_ARGS ?=
 PROTOC := $(BIN)/protoc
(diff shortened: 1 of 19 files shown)


jhump:
cc: @drice-buf (just an FYI; no need to review if you are busy)

Makefile
```diff
@@ -43,34 +43,34 @@ clean: ## Cleans the underlying build.
 runconformance: generate $(CONNECT_CONFORMANCE) ## Run the new conformance test suite.
 	./gradlew $(GRADLE_ARGS) conformance:client:google-java:installDist conformance:client:google-javalite:installDist
 	$(CONNECT_CONFORMANCE) -v --mode client --conf conformance/client/lite-unary-config.yaml \
-		--known-failing @conformance/client/known-failing-unary-cases.txt -- \
+		--known-failing @conformance/client/known-failing-unary-cases.txt --trace -- \
```
pkwarren, line 46: Perhaps we could set `CONNECT_CONFORMANCE_ARGS ?= -v --mode client --trace` which would allow easily running with additional options if needed.

pkwarren (approved):
Changes look great - the updates to ConnectException/errorDetailsParser are a big improvement.

Makefile
jhump: Ah, nice. I like it.

Done.

PR #274, opened 2024-05-15
Fix all conformance failures other than timeouts/deadlines

jhump:
The first two commits include fixes to the protocol logic to fix the known failures in conformance tests.

This still leaves the timeout/deadline propagation cases failing since those will require something more significant to resolve.

* The first commit fixes how trailers-only responses are classified. This was previously just looking for a "grpc-status" key in the headers. If it was present, it was treating it as a trailers-only response, even if there was a body and/or trailers.
* The second commit fixes the client to report errors in the face of unexpected response content types. I wish this were less verbose: I think in the future we can likely find some low-hanging fruit ways to improve the code organization and increase code sharing in these protocol implementations (especially gRPC and gRPC-Web).

The third commit is just some minor updates to comments and TODOs that I figured I'd tweak while I was in these files. Hopefully they all make sense.

The final commit cleans up the known failing test case configuration -- to remove everything but the timeout/deadline tests.

### DIFF
--- a/library/src/main/kotlin/com/connectrpc/protocols/ErrorJSONModels.kt
+++ b/library/src/main/kotlin/com/connectrpc/protocols/ErrorJSONModels.kt
@@ -38,5 +38,9 @@ internal class EndStreamResponseJSON(
 )
 
 internal fun contentTypeIsJSON(contentType: String): Boolean {
+    // TODO: This could be more robust, like actually parsing the content-type.
+    // There exists a good helper for that, but it's in okhttp, which we intentionally
+    // don't have as a dep for this module, which aims to be agnostic of the actual
+    // HTTP client implementation to use.
     return contentType == "application/json" || contentType == "application/json; charset=utf-8"
 }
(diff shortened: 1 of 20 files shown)


jhump:
cc @perezd

library/src/main/kotlin/com/connectrpc/protocols/ErrorJSONModels.kt
```diff
@@ -36,3 +36,7 @@ internal class EndStreamResponseJSON(
     @Json(name = "error") val error: ErrorPayloadJSON?,
     @Json(name = "metadata") val metadata: Headers?,
 )
+
+internal fun contentTypeIsJSON(contentType: String): Boolean {
+    return contentType == "application/json" || contentType == "application/json; charset=utf-8"
```
pkwarren, line 41: Should we use `MediaType` to parse these reliably (we use this in `ConnectOkHttpClient.kt`)?

pkwarren (approved):
Looks great - just one suggestion.

pkwarren (approved):

PR #276, opened 2024-05-17
Add timeout enforcement to ProtocolClient

### LINKED ISSUE #249: Timeout and deadline propagation issues
Currently, this library relies wholly on the underlying `HTTPClientInterface` implementation handling timeouts. That poses two problems:
1. The only implementation of `HTTPClientInterface` provided is based on `okhttp`, which does not enforce timeouts on full-duplex bidirectional operations after the stream is established.
2. It should not be up to this layer to handle deadline propagation -- communicating the timeout to the server, so it can attenuate the deadline as it processes the request. That is a protocol concern (and Connect and gRPC/gRPC-Web both handle it slightly differently). Currently, timeout headers are never sent, so this client does not attempt to propagate deadlines to the server.

So to address these deficiencies:
1. There needs to be a mechanism for configuring timeouts _other than_ the `HTTPClientInterface` implementation. This could be in the form of default timeouts in `ProtocolClientConfig`; it could be some way to specify timeouts on a per-RPC basis; or it could be both.
2. When such a timeout is configured, the protocol handling code needs to encode the timeout in a request header to communicate the deadline to the server.

The underlying `HTTPClientInterface` might still impose its own timeouts. So it's an open question as to whether this needs to be considered in the timeout handling. For example, if the user sets a 20 second timeout for an RPC, but the underlying HTTP client has a 10 second timeout configured, should the library take this into consideration and communicate a 10 second deadline to the server? If so, then we may need to add methods to `HTTPClientInterface` (or create sub-interfaces that can be optionally implemented) so that the framework can ask the client about its timeouts, in order to correctly consider them.

Alternatively, instead of the framework asking the client about its timeout, it could _tell_ the client what timeout to use (similarly to how #14 proposes that the framework may need to tell the client what protocols to support, depending on the RPC protocol and URL scheme used for a particular operation).

jhump:
This moves the timeout enforcement out of the HTTPClientInterface and into the ProtocolClientInterface. This is more appropriate so that the protocol implementations can be aware of the timeout and add a timeout header, to propagate the deadline to the server.

I've broken it up into a sequence of commits that hopefully makes it easier to review.

I tried looking into making the timeout scheduler coroutine-aware. But this caused issues. Even though I could propagate the `CoroutineScope` from the caller, when invoked from a unary suspend function or a stream function, the issue is how `CoroutineScope.launch` impacts the scope lifetime. In particular, a scope can't complete until all of its children complete, so I was running into issues where code was incorrectly blocking on that child coroutine (which was just doing `delay` for the full timeout period) to end.

I didn't debug it enough to get to bedrock regarding where the blocking was happening or exactly why (naively, I would have expected the timeout coroutine to just be a part of the main coroutine scope create in the initial `runBlocking` in the conformance client, which shouldn't cause the blocking issues I observed 🤷).

So using a simpler Timer seemed like the next best thing. And an alternative like this was needed anyway to be called from non-suspend functions like the callback and blocking unary signatures (for which there is no parent coroutine scope in which to launch a new coroutine).

Resolves #249.

### DIFF
--- a/library/src/main/kotlin/com/connectrpc/ProtocolClientConfig.kt
+++ b/library/src/main/kotlin/com/connectrpc/ProtocolClientConfig.kt
@@ -25,6 +25,8 @@ import com.connectrpc.protocols.NetworkProtocol
 import java.net.URI
 import kotlin.coroutines.CoroutineContext
 import kotlin.time.Duration
+import kotlin.time.DurationUnit
+import kotlin.time.toDuration
 
 typealias TimeoutOracle = (MethodSpec<*, *>) -> Duration?
 
@@ -81,7 +83,10 @@ class ProtocolClientConfig @JvmOverloads constructor(
     // is returned, the entire call must complete before it elapses. If the
     // call is still active at the end of the timeout period, it is cancelled
     // and will result in an exception with a Code.DEADLINE_EXCEEDED code.
-    val timeoutOracle: TimeoutOracle = { null },
+    //
+    // The default oracle, if not configured, returns a 10 second timeout for
+    // all operations.
+    val timeoutOracle: TimeoutOracle = { 10.toDuration(DurationUnit.SECONDS) },
     // Schedules timeout actions.
     val timeoutScheduler: Timeout.Scheduler = Timeout.DEFAULT_SCHEDULER,
 ) {
--- a/library/src/main/kotlin/com/connectrpc/ProtocolClientConfig.kt
+++ b/library/src/main/kotlin/com/connectrpc/ProtocolClientConfig.kt
@@ -16,6 +16,7 @@ package com.connectrpc
 
 import com.connectrpc.compression.CompressionPool
 import com.connectrpc.compression.GzipCompressionPool
+import com.connectrpc.http.Timeout
 import com.connectrpc.protocols.ConnectInterceptor
 import com.connectrpc.protocols.GETConfiguration
 import com.connectrpc.protocols.GRPCInterceptor
@@ -23,11 +24,32 @@ import com.connectrpc.protocols.GRPCWebInterceptor
 import com.connectrpc.protocols.NetworkProtocol
 import java.net.URI
 import kotlin.coroutines.CoroutineContext
+import kotlin.time.Duration
+
+typealias TimeoutOracle = (MethodSpec<*, *>) -> Duration?
+
+/**
+ * Returns an oracle that provides the given timeouts for unary or stream
+ * operations, respectively.
+ */
+fun simpleTimeouts(unaryTimeout: Duration?, streamTimeout: Duration?): TimeoutOracle {
+    return { methodSpec ->
+        when (methodSpec.streamType) {
+            StreamType.UNARY -> unaryTimeout
+            else -> streamTimeout
+        }
+    }
+}
 
 /**
  *  Set of configuration used to set up clients.
  */
 class ProtocolClientConfig @JvmOverloads constructor(
+    // TODO: Use a block-based construction pattern instead of JvmOverloads
+    //       so we can add new fields in the future without having to worry
+    //       about their ordering or potentially breaking compatibility with
+    //       already-compiled byte code.
+
     // The host (e.g., https://connectrpc.com).
     val host: String,
     // The client to use for performing requests.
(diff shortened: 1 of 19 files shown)


library/src/main/kotlin/com/connectrpc/ProtocolClientConfig.kt
```diff
@@ -81,7 +83,10 @@ class ProtocolClientConfig @JvmOverloads constructor(
     // is returned, the entire call must complete before it elapses. If the
     // call is still active at the end of the timeout period, it is cancelled
     // and will result in an exception with a Code.DEADLINE_EXCEEDED code.
-    val timeoutOracle: TimeoutOracle = { null },
+    //
+    // The default oracle, if not configured, returns a 10 second timeout for
+    // all operations.
+    val timeoutOracle: TimeoutOracle = { 10.toDuration(DurationUnit.SECONDS) },
```
jhump, line 89: I added this to account for _removal_ of the timeouts in the configuration helper below. So if users use `ConnectOkHttpClient.configureClient` and forget to set timeouts in this config, they get the same default behavior as the `OkHttpClient` was providing (except that this default applies to bidirectional streams, whereas the `OkHttpClient` timeouts do not).

jhump:
@pkwarren, ping.

library/src/main/kotlin/com/connectrpc/http/Timeout.kt
```diff
@@ -0,0 +1,86 @@
...
+
+    companion object {
+        /**
+         * A default implementation that a Timer backed by a single daemon thread.
+         * The thread isn't started until the first cancelation is scheduled.
+         */
+        val DEFAULT_SCHEDULER = object : Scheduler {
+            override fun scheduleTimeout(delay: Duration, action: Cancelable): Timeout {
+                val timeout = Timeout(action)
+                val task = timerTask { timeout.trigger() }
```
pkwarren, line 77: This works but we might consider tradeoffs vs. ScheduledThreadPoolExecutor. Probably not a big deal for timeout scheduling.

okhttp/src/main/kotlin/com/connectrpc/okhttp/ConnectOkHttpClient.kt
```diff
@@ -43,24 +43,38 @@ import okio.BufferedSink
...
+import java.time.Duration
 
 /**
  * The OkHttp implementation of HTTPClientInterface.
  */
 class ConnectOkHttpClient @JvmOverloads constructor(
-    unaryClient: OkHttpClient = OkHttpClient(),
-    streamClient: OkHttpClient = unaryClient,
+    private val unaryClient: OkHttpClient = OkHttpClient(),
+    // TODO: remove this; a separate client is only useful for configuring separate timeouts,
```
pkwarren, line 53: This is not just for timeouts IIRC - you may also want to configure pings on clients for long lived streams but not on ones for unary calls.

pkwarren: Here's some background on the two clients (aside from timeouts which will be much better behaved with this PR): https://github.com/connectrpc/connect-kotlin/issues/13

jhump: It's unclear why you'd want PINGs on one client and not the other. If you have created a connection that uses PINGs, just use that same client for both unary and stream traffic. I don't see any downside to that. It wouldn't increase the PING activity since you're not creating two clients that both use PING -- but instead using the PINGing client for all RPCs.

As far as the read timeout being different between unary and streams, that is now even more configurable (based on `MethodSpec`) via the timeout oracle. So you can set the read timeout to zero (or use `ConnectOkHttpClient.configureClient`) and it should be appropriate for both unary and stream traffic.

@kohenkatz, please take a look at this thread and PR and let me know what you think. Note that this PR is not removing the second client -- not yet. But I don't think it will be needed anymore and would appreciate your input, like if you see other reasons to keep the second client in future versions.

pkwarren (approved):

okhttp/src/main/kotlin/com/connectrpc/okhttp/ConnectOkHttpClient.kt
kohenkatz: @jhump I'm not sure how I missed the notification of this comment 10 months ago ... 😳.

I suspect that you are right that we don't really need the second client anymore, but I'm going to look at that in a little more detail this week.

PR #312, opened 2024-08-05
Determine protoc version from protobuf version

pkwarren:
The protoc version is the same value as the '[versions] protobuf' in the Gradle version catalog, without the major version. Update the Makefile to automatically keep these versions in sync, removing a manual step of updating the Makefile whenever dependabot updates protobuf.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,10 @@ CACHE := .tmp/cache
 LICENSE_HEADER_YEAR_RANGE := 2022-2023
 LICENSE_HEADER_VERSION := v1.34.0
 CONFORMANCE_VERSION := v1.0.2
-PROTOC_VERSION ?= 27.2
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -11,7 +11,10 @@ CACHE := .tmp/cache
 LICENSE_HEADER_YEAR_RANGE := 2022-2023
 LICENSE_HEADER_VERSION := v1.34.0
 CONFORMANCE_VERSION := v1.0.2
-PROTOC_VERSION ?= 27.2
+PROTOC_VERSION ?= $(shell yq '.versions.protobuf' gradle/libs.versions.toml | cut -d'.' -f2-)
```
jhump, line 14: Whoah. I didn't realize `yq` understood toml files.

jhump (approved):

Issue #333, opened 2024-10-01
Generate a @kotlin.Deprecated annotation for deprecated RPCs


timostamm:
Protobuf has a built-in option to mark elements as deprecated. For example:

```proto
syntax="proto3";

service FooService {
  rpc Foo(FooRequest) returns (FooResponse) {
    // Is this method deprecated?
    // Depending on the target platform, this can emit Deprecated annotations
    // for the method, or it will be completely ignored; in the very least,
    // this is a formalization for deprecating methods.
    option deprecated = true;
  }
}

message FooRequest {}
message FooResponse {}
```

The gRPC implementation generates a `kotlin.Deprecated` annotation for this RPC:

```kotlin
    @Deprecated("The underlying service method is marked deprecated.")
    public suspend fun foo(request: FooRequest, headers: Metadata = Metadata()): FooResponse = ...
```

The feature was added in https://github.com/grpc/grpc-kotlin/pull/264

Since it can be useful feature for team collaboration, I think we should do the same in the plugin for Connect. 

For the example above, the generated client would change as follows:

```diff
public class FooServiceClient(
  private val client: ProtocolClientInterface,
) : FooServiceClientInterface {
+ @kotlin.Deprecated("The underlying service method is marked deprecated.")
  override suspend fun foo(request: FooRequest, headers: Headers): ResponseMessage<FooResponse> =
```

The generated client interface would change as well:

```diff
public interface FooServiceClientInterface {
+ @kotlin.Deprecated("The underlying service method is marked deprecated.")
  public suspend fun foo(request: FooRequest, headers: Headers = emptyMap()):
```

PR #334, opened 2024-10-02
Generate @Deprecated annotations

timostamm:
This updates the plugin `protoc-gen-connect-kotlin` to generate annotations for deprecated elements. Closes https://github.com/connectrpc/connect-kotlin/issues/333.

If an RPC is marked deprecated with the option `deprecated = true`, the corresponding member of the generated client interface and implementation is annotated with:

```kotlin
@Deprecated("The method is deprecated in the Protobuf source file.")
```

Deprecating entire services is supported as well. If `deprecated = true` is set on a service, the generated client interface and implementation is annotated with:

```kotlin
@Deprecated("The service is deprecated in the Protobuf source file.")
```

For good measure, deprecating Protobuf files is supported as well. Because `kotlin.Deprecated` is not applicable to files, we annotate every client interface and implementation generated from the file:

```kotlin
@Deprecated("The Protobuf source file that defines this service is deprecated.")
```


protoc-gen-connect-kotlin/src/main/kotlin/com/connectrpc/protocgen/connect/Generator.kt
```diff
@@ -524,3 +539,36 @@ private fun String.packageToDirectory(): String {
...
 }
+
+private fun TypeSpec.Builder.addServiceDeprecation(
+    service: Descriptors.ServiceDescriptor,
+    file: Descriptors.FileDescriptor
+): TypeSpec.Builder {
+    if (service.options.deprecated) {
+        this.addAnnotation(
+            AnnotationSpec.builder(Deprecated::class)
+                .addMember("%S", "The underlying service is marked deprecated.")
```
jhump, line 550: Maybe instead of "underlying" say "Protobuf source file"?
```suggestion
                .addMember("%S", "The service is deprecated in the Protobuf source file.")
```
WDYT?

protoc-gen-connect-kotlin/src/main/kotlin/com/connectrpc/protocgen/connect/Generator.kt
```diff
@@ -524,3 +539,36 @@ private fun String.packageToDirectory(): String {
...
+                .build()
+        )
+    } else if (file.options.deprecated) {
+        this.addAnnotation(
+            AnnotationSpec.builder(Deprecated::class)
+                .addMember("%S", "The underlying file is marked deprecated.")
```
jhump, line 556: Ditto. This is the main one that I think might be confusing since it could be very unclear to a reader what "underlying file" means.
```suggestion
                .addMember("%S", "The Protobuf source file that defines this service is deprecated.")
```

protoc-gen-connect-kotlin/src/main/kotlin/com/connectrpc/protocgen/connect/Generator.kt
```diff
@@ -524,3 +539,36 @@ private fun String.packageToDirectory(): String {
...
+    return this
+}
+
+private fun FunSpec.Builder.addMethodDeprecation(
+    method: Descriptors.MethodDescriptor,
+): FunSpec.Builder {
+    if (method.options.deprecated) {
+        this.addAnnotation(
+            AnnotationSpec.builder(Deprecated::class)
+                .addMember("%S", "The underlying service method is marked deprecated.")
```
jhump, line 569:
```suggestion
                .addMember("%S", "The method is deprecated in the Protobuf source file.")
```

jhump (approved):
LGTM. Small (non-blocking) suggestion on deprecation message wording. Feel free to accept suggestions or not.

protoc-gen-connect-kotlin/src/main/kotlin/com/connectrpc/protocgen/connect/Generator.kt
timostamm: The idea was to match gRPC, but I agree that "underlying" is not clear. Updated in https://github.com/connectrpc/connect-kotlin/pull/334/commits/1a37603b993b3039f6156ddeda439569a3e634c1.

erawhctim:
@timostamm thanks for tackling this! 

One question: if a comment exists on the deprecation option in the proto file, or on an rpc definition, would it be feasible to surface that comment as part of the Deprecated annotation in the generated Kotlin code? 

Ideally, the proto spec itself would support some kind of deprecation message field to define this in a more official way, but I can very easily imagine our developers adding a comment to explain the deprecation and/or point to a replacement, non-deprecated field/rpc call. If there's a way we can include that context in Kotlin, that would be a huge win. Is that possible with what we have today?

timostamm:
> Looks like you may need to add the "ignore deprecations" annotation at the top of the generated files, like the base protoc plugins do, to fix the CI build issue.

Generating `@file:Suppress("DEPRECATION")` allows to compile the file without warning, but using the deprecated code elsewhere will still raise a warning. This works out well. Done in https://github.com/connectrpc/connect-kotlin/pull/334/commits/eec519e293503a0a7fe11e31871a6e5099e1f8d1.

timostamm:
@erawhctim, I don't even know a language from the top of my head that supports deprecations, but doesn't support deprecation messages. Unfortunately, the option for deprecating elements in Protobuf is typed as a boolean ([source](https://github.com/protocolbuffers/protobuf/blob/v28.2/src/google/protobuf/descriptor.proto#L883-L887)), and doesn't allow users or us to provide a description.

I think the most reasonable approach for now is to add a comment in the Protobuf source when you set `deprecated = true`. The comments will show up as KDoc.

jhump (approved):

jhump:
> would it be feasible to surface that comment as part of the Deprecated annotation in the generated Kotlin code?

Unfortunately, plugins do not have access to such comments.

Plugins must inspect a file descriptor's "source code info" to find comments. While "spans" (the location of an element in the source file) are generated for nearly everything, comments are only retained for complete declarations. The field itself is a complete declaration (so it's leading and trailing comments are preserved). But comments for elements _inside_ the field -- like individual options inside the `[...]` -- are not retained in source code info and thus not available to plugins.

PR #528, opened 2026-03-17
Fix CI and release workflow warnings

pkwarren:
Update to the newer buf-action and gradle actions. Pin all third party actions. Remove old multidex dependency not needed in newer android versions.


jhump (commented):
Seems fine to me, but I was a little confused at the use of a sha instead of a major version for pinning the buf-action.

timostamm (approved):

PR #555, opened 2026-06-12
Update compile/target sdk and fix AGP warnings

pkwarren:
Newer dependencies (like okhttp) require that we compile to a later SDK. Fix two AGP warnings for deprecated settings which will go away in the next release.

### DIFF
--- a/examples/android/build.gradle.kts
+++ b/examples/android/build.gradle.kts
@@ -3,12 +3,12 @@ plugins {
 }
 
 android {
-    compileSdk = 35
+    compileSdk = 36
 
     defaultConfig {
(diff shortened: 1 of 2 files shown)


examples/android/build.gradle.kts
```diff
@@ -3,12 +3,12 @@ plugins {
 }
 
 android {
-    compileSdk = 35
+    compileSdk = 36
```
pkwarren, line 6: This will fix the build error here: https://github.com/connectrpc/connect-kotlin/pull/554

jhump (approved):

PR #573, opened 2026-07-27
Upgrade to kotlin 2.4 compiler

pkwarren:
Update to the kotlin 2.4 compiler, and bump API/language version to 2.2 (keeping with Kotlin's support for the last N-2 releases). Fix warnings from the upgrade.

### DIFF
--- a/library/src/main/kotlin/com/connectrpc/protocols/ErrorJSONModels.kt
+++ b/library/src/main/kotlin/com/connectrpc/protocols/ErrorJSONModels.kt
@@ -15,26 +15,25 @@
 package com.connectrpc.protocols
 
 import com.connectrpc.Headers
-import com.squareup.moshi.Json
 import com.squareup.moshi.JsonClass
 
 @JsonClass(generateAdapter = true)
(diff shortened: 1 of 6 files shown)


library/src/main/kotlin/com/connectrpc/protocols/ErrorJSONModels.kt
```diff
@@ -15,26 +15,25 @@
...
 import com.squareup.moshi.JsonClass
 
 @JsonClass(generateAdapter = true)
 internal class ErrorPayloadJSON(
-    @Json(name = "code") val code: String?,
-    @Json(name = "message") val message: String?,
-    @Json(name = "details") val details: List<ErrorDetailPayloadJSON>?,
+    val code: String?,
+    val message: String?,
+    val details: List<ErrorDetailPayloadJSON>?,
```
pkwarren, line 24: The annotations triggered a warning unless we changed to `@param:Json` but since the field names matched the annotations everywhere, we didn't need the annotations at all so I removed them.

timostamm (approved):

